PR #7, opened 2018-02-02
TCP onion packets

### LINKED ISSUE #4: Tcp relay onion packets
Implement packet parsing as it was done in https://github.com/zetok/tox/blob/master/src/toxcore/tcp/packet.rs

kurnevsky:
Closes #4

### DIFF
--- a/src/toxcore/tcp/packet.rs
+++ b/src/toxcore/tcp/packet.rs
@@ -26,7 +26,12 @@
 use toxcore::binary_io_new::*;
 use toxcore::crypto_core::*;
 
-use nom::{be_u8, be_u16, be_u64, rest};
+use nom::{be_u8, be_u16, be_u64, le_u8, rest};
+use std::net::{
+    IpAddr,
+    Ipv4Addr,
+    Ipv6Addr,
+};
 
 /** Top-level TCP packet.
 
@@ -50,10 +55,10 @@ pub enum Packet {
     OobSend(OobSend),
     /// [`OobReceive`](./struct.OobReceive.html) structure.
     OobReceive(OobReceive),
-    /// TODO
-    //OnionDataRequest,
-    /// TODO
-    //OnionDataResponse,
+    /// [`OnionRequest`](./struct.OnionRequest.html) structure.
+    OnionRequest(OnionRequest),
+    /// [`OnionResponse`](./struct.OnionResponse.html) structure.
+    OnionResponse(OnionResponse),
     /// [`Data`](./struct.Data.html) structure.
     Data(Data)
 }
@@ -71,6 +76,8 @@ impl FromBytes for Packet {
         map!(PongResponse::from_bytes, Packet::PongResponse) |
         map!(OobSend::from_bytes, Packet::OobSend) |
         map!(OobReceive::from_bytes, Packet::OobReceive) |
+        map!(OnionRequest::from_bytes, Packet::OnionRequest) |
+        map!(OnionResponse::from_bytes, Packet::OnionResponse) |
         map!(Data::from_bytes, Packet::Data)
     ));
 }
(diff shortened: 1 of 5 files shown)


src/toxcore/tcp/packet.rs
```diff
@@ -457,6 +466,168 @@ impl ToBytes for OobReceive {
     }
 }
 
+
+
+
+
+
+
+// TODO: move it somewhere else
```
kurnevsky: Unify with DHT.

src/toxcore/tcp/packet.rs
```diff
@@ -457,6 +466,168 @@ impl ToBytes for OobReceive {
...
+pub struct IpPort {
+    pub addr: IpAddr,
+    pub port: u16
+}
+
+impl FromBytes for IpPort {
+    named!(from_bytes<IpPort>, do_parse!(
+        ip_type: le_u8 >>
+        addr: alt!(
+            cond_reduce!(ip_type == 2 || ip_type == 130, terminated!(
```
kpp: In TCP Relay we must strictly accept only TCP4 and TCP6

kurnevsky: Not exactly. They have bool flag disable_family_check (see onion.c#65) and set it to true when parse IP for TCP server. So TCP server can accept both TCP and UDP families.

src/toxcore/tcp/server/server.rs
```diff
@@ -672,11 +673,50 @@ mod tests {
         ));
     }
     #[test]
+    fn handle_onion_request() {
+        let (tcp_onion_sink, tcp_onion_stream) = mpsc::unbounded();
+        let server = Server::new(tcp_onion_sink);
+        let (client_pk, _, _) = add_random_client(&server);
```
kpp: I don't like these `_, _`. Please replace it with `_ip_addr, _rx....`

src/toxcore/tcp/server/server.rs
```diff
@@ -603,10 +673,50 @@ mod tests {
...
+            data: [13; 100].to_vec()
+        };
+        let handle_res = server
+            .handle_packet(&client_pk, Packet::OnionRequest(request.clone()))
+            .wait();
+        assert!(handle_res.is_ok());
+
+        let (packet, _) = tcp_onion_stream.into_future().wait().unwrap();
+        assert_eq!(packet.unwrap(), request);
+    }
```
kpp: Add `handle_onion_response` and check error. See https://coveralls.io/builds/15364599/source?filename=src%2Ftoxcore%2Ftcp%2Fserver%2Fserver.rs#L89 for missing lines.

kpp:
See test for server.rs `send_anything_to_dropped_client` how to test dropped onion.

src/toxcore/tcp/server/server.rs
```diff
@@ -323,37 +368,56 @@ impl Server {
 
 #[cfg(test)]
 mod tests {
+    extern crate rand;
```
kpp: Oh cmon. These kind of tests do not require randomness.

kurnevsky: Yes, but I added it for method 'add_random_client'. In other case I had to pass IpAddr there every time I need to add client because this method is used to add multiple clients. I can remove randomness here but it will increase boilerplate.

kurnevsky: And perhaps in this case we should get rid of gen_keypair() as well :)

kpp: There is a non 0 probability that 2 calls for a rand addr will return **the same addr**. And for now the code is unspecified what will happen if the client connects with the same addr.

PR #9, opened 2018-02-06
fix(tcp_server): find client by ip with port

kurnevsky:


src/toxcore/tcp/server/server.rs
```diff
@@ -50,7 +50,7 @@ to clients via network.
 #[derive(Clone)]
 pub struct Server {
     connected_clients: Rc<RefCell<HashMap<PublicKey, Client>>>,
-    keys_by_ip_addr: Rc<RefCell<HashMap<IpAddr, PublicKey>>>,
+    keys_by_addr: Rc<RefCell<HashMap<(IpAddr, u16), PublicKey>>>,
```
kpp: `(IpAddr, u16)` -> `(IpAddr, /*port*/ u16)`

PR #15, opened 2018-02-11
Add dht_node_new.rs

NamsooCho:
Add new example code for dht_node.
It rely on currently under constructing local dht_node which is not yet pushed.
So this code do not work on current master branch.
dht_node.rs pairing with this example code will be pushed later.


NamsooCho:
After discussion,
this PR will be closed.

kpp:
This PR is closed in favor of https://github.com/tox-rs/tox/pull/25

PR #19, opened 2018-02-17
refactor(dht): move io traits impl for ip types to toxcore

kurnevsky:


src/toxcore/binary_io_new.rs
```diff
@@ -46,3 +53,59 @@ impl FromBytes for PublicKey {
 impl FromBytes for Nonce {
     named!(from_bytes<Nonce>, map_opt!(take!(NONCEBYTES), Nonce::from_slice));
 }
+
+// Ip Type
+// Value | Type
+// ----- | ----
+// `2`   | UDP IPv4
```
kpp: Really? `IpAddr` Does not contain the type.

kurnevsky: ok, removed this comment - it was present in original code.

PR #23, opened 2018-02-22
Tests and more

kurnevsky:


src/toxcore/binary_io_new.rs
```diff
@@ -102,3 +102,20 @@ impl ToBytes for Ipv6Addr {
...
+#[cfg(test)]
+macro_rules! encode_decode_test (
+    ($test:ident, $value:expr) => (
+        #[test]
+        fn $test() {
+            let value = $value;
+            let mut buf = [0; 1024];
+            let (_, size) = value.to_bytes((&mut buf, 0)).unwrap();
+            let (rest, decoded_value) = FromBytes::from_bytes(&buf[..size]).unwrap();
+            fn infer<T>(_: &T, _: &T) { }
```
kpp: what is it for?

src/toxcore/binary_io_new.rs
```diff
@@ -102,3 +102,20 @@ impl ToBytes for Ipv6Addr {
...
 }
+
+#[cfg(test)]
+macro_rules! encode_decode_test (
+    ($test:ident, $value:expr) => (
+        #[test]
+        fn $test() {
+            let value = $value;
+            let mut buf = [0; 1024];
+            let (_, size) = value.to_bytes((&mut buf, 0)).unwrap();
```
kpp: assert size <= 1024

src/toxcore/binary_io_new.rs
kurnevsky: Without this compiler can't infer type of decoded_value, it seems assert_eq! doesn't give enough information so this is workaround to say that decoded_value has the same type as value.

PR #25, opened 2018-02-26
Refactor DHT node

NamsooCho:


src/toxcore/dht_new/client.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// respond with NatPingResponse to peer
+    pub fn send_nat_ping_response(&self, rpk: &PublicKey, resp_payload: NatPingResponse) -> IoFuture<()> {
+        let payload = DhtRequestPayload::NatPingResponse(resp_payload);
+        let nat_ping_resp = DhtPacket::DhtRequest(DhtRequest::new(&self.precomputed_key.clone(), rpk, &self.pk, payload));
+        self.send(nat_ping_resp)
+    }
+    /// relay NatPingRequest/NatPingResponse to target peer
+    pub fn relay_nat_ping_packet(&self, addr: &SocketAddr, request: DhtRequest) -> IoFuture<()> {
+        let packet = DhtPacket::DhtRequest(request);
+        Box::new(self.tx.clone() // clone tx sender for 1 send only
```
kpp: Create method `send_to` as method `send` but with explicit addr-argument. Method `send` should use `send_to` inside it passing `self.addr` as addr.

examples/dht_node_new.rs
```diff
@@ -0,0 +1,124 @@
...
+    // Create a channel for this socket
+    let (tx_sink, rx_sink) = mpsc::unbounded::<DhtUdpPacket>();
+    // The server task asynchronously iterates over and processes each
+    // incoming packet.
+    let handler = stream.for_each(move |(addr, packet)| {
+        let tx_sink_c = tx_sink.clone();
+        // Create a channel for this peer
+        let (tx, rx) = mpsc::unbounded();
+        // Add or update an entry for this `Peer` in the connected client hashmap.
+        let client = node.borrow_mut().add_client(&addr, &packet, tx);
```
kpp: replace `tx` with `tx_sink_c`. You don't need to create `let (tx, rx) = mpsc::unbounded();`, pass `tx` into `add_client`, create `let writer = rx` future. Just use `tx_sink_c`.

examples/dht_node_new.rs
```diff
@@ -0,0 +1,124 @@
...
+        error!("packet receive error = {:?}", err);
+        Error::new(ErrorKind::Other, "udp receive error")
+    });
+
+    let writer_timer = tokio_timer::wheel()
+        .tick_duration(time::Duration::from_secs(1))
+        .build()
+    ;
+
+    let relay = rx_sink
```
kpp: rename `relay` -> `writer`

src/toxcore/dht_new/dht_node.rs
```diff
@@ -0,0 +1,528 @@
...
+        Ok(DhtNode {
+            sk: sk,
+            pk: pk,
+            kbucket: kbucket,
+            connected_clients: HashMap::new(),
+        })
+    }
+
+    /// add new client entry in connected_clients hashmap
+    pub fn add_client(&mut self, addr: &SocketAddr, packet: &Option<DhtPacket>, tx: Tx) -> Client {
```
kpp: Do you actually need to add client? To create it? It worth to keep clients in tcp server, because TCP connection is stateful and we actually need to keep `tx` for each client to send the data into appropriate TCP connection linked to this client. On the other hand, we do have 1 UDP connection and we send data through it with the only one difference: `addr` arg.

src/toxcore/dht_new/dht_node.rs
```diff
@@ -0,0 +1,528 @@
...
+            _ => unreachable!("Received packet is not a DhtPacket"),
+        };
+        let client_result = 
+            if let Some(mut client) = self.connected_clients.get_mut(&pk) {
+                if client.addr != *addr {
+                    client.addr = *addr;
+                }
+                client.clone()
+            } else {
+                let precomputed_key = encrypt_precompute(&pk, &self.sk);
```
kpp: Calling `encrypt_precompute` only once is a good performance optimization, but do we really need this optimization for now? We may do this optimization next PR

examples/dht_node_new.rs
```diff
@@ -0,0 +1,124 @@
...
+    // The server task asynchronously iterates over and processes each
+    // incoming packet.
+    let handler = stream.for_each(move |(addr, packet)| {
+        let tx_sink_c = tx_sink.clone();
+        // Create a channel for this peer
+        let (tx, rx) = mpsc::unbounded();
+        // Add or update an entry for this `Peer` in the connected client hashmap.
+        let client = node.borrow_mut().add_client(&addr, &packet, tx);
+
+        let _ = node.borrow_mut().handle_packet(&(addr, packet.unwrap()), client);
```
kpp: Why do you pass `&(addr, packet)` by ref? Try to pass values by value.

src/toxcore/dht_new/dht_node.rs
```diff
@@ -0,0 +1,528 @@
...
+
+    /// add new client entry in connected_clients hashmap
+    pub fn add_client(&mut self, addr: &SocketAddr, packet: &Option<DhtPacket>, tx: Tx) -> Client {
+        let pk = match packet.clone().unwrap() {
+            DhtPacket::PingRequest(ping_request) => ping_request.pk,
+            DhtPacket::PingResponse(ping_response) => ping_response.pk,
+            DhtPacket::NodesRequest(nodes_request) => nodes_request.pk,
+            DhtPacket::NodesResponse(nodes_response) => nodes_response.pk,
+            DhtPacket::DhtRequest(dht_request) => dht_request.spk,
+            _ => unreachable!("Received packet is not a DhtPacket"),
```
kpp: Also `add_client` must not accept packet to get `pk`. You may create a method to extract client pk from DhtPacket, call it before `add_client` and pass `pk` into `add_client` as computed argument.

src/toxcore/dht_new/mod.rs
```diff
@@ -27,3 +27,5 @@ pub mod packet;
 pub mod kbucket;
 pub mod packed_node;
 pub mod codec;
+pub mod dht_node;
```
kpp: Does it have to be named as `dht_node` while the next line is `client`? Shouldn't it be `server`?

src/toxcore/dht_new/dht_node.rs
NamsooCho: Good catch.
I was missing it.
UDP connection doesn't need to be saved.

src/toxcore/dht_new/client.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// peer info.
+#[derive(Clone, Debug)]
+pub struct Client {
+    /// Public key of peer
+    pub pk: PublicKey,
+    /// socket address of peer
+    pub addr: SocketAddr,
+    /// tx split of channel to send packet to this peer via udp socket
+    pub tx: Tx,
```
kurnevsky: I would say the server struct should store tx but not the client since it's shared between all clients. And the same for send/send_to methods.

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,525 @@
...
+            let pk = Server::get_pk(&nat_ping_req);
+            let client = alice.create_client(&addr.clone(), pk, tx.clone());
+            alice.handle_packet((addr, nat_ping_req), client).wait();
+            let (packet, rx) = rx.into_future().wait().unwrap();
+            debug!("received packet {:?}", packet);
+            if let DhtPacket::DhtRequest(packet) = packet.unwrap().1 {
+                if let DhtRequestPayload::NatPingResponse(nat_ping_resp_payload) = packet.get_payload(&alice.sk).unwrap() {
+                    assert_eq!(nat_ping_resp_payload.id, nat_req.id);
+                } else {
+                    panic!("can not occur");
```
kurnevsky: unreachable! :)

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,525 @@
...
+
+    /// get pk of DhtPacket
+    pub fn get_pk(packet: &DhtPacket) -> PublicKey {
+        match packet.clone() {
+            DhtPacket::PingRequest(ping_request) => ping_request.pk,
+            DhtPacket::PingResponse(ping_response) => ping_response.pk,
+            DhtPacket::NodesRequest(nodes_request) => nodes_request.pk,
+            DhtPacket::NodesResponse(nodes_response) => nodes_response.pk,
+            DhtPacket::DhtRequest(dht_request) => dht_request.spk,
+            _ => unreachable!("Received packet is not a DhtPacket"),
```
kurnevsky: Actually it's not unreachable because it depends on how this method is called, so panic is preferable in such cases.

kurnevsky: But why we can't just get pk when we need it from the packet directly?

examples/dht_node_new.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    let (sink, stream) = listener.framed(DhtCodec).split();
+    // Create a channel for this socket
+    let (tx, rx) = mpsc::unbounded::<DhtUdpPacket>();
+    // The server task asynchronously iterates over and processes each
+    // incoming packet.
+    let handler = stream.for_each(move |(addr, packet)| {
+        let tx_c = tx.clone();
+        let pk = Server::get_pk(&packet.clone().unwrap());
+        let client = node.borrow_mut().create_client(&addr, pk, tx_c);
```
kurnevsky: It seems to me that server should take only packet and maybe socket address but not the whole client. It should manage clients inside so the user of server api wouldn't even know about it.

src/toxcore/dht_new/server.rs
kpp: > But why we can't just get pk when we need it from the packet directly?

How?

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,525 @@
...
+            }
+            else { // do nothing
+                Box::new( future::ok(()) )
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
kpp: `mod tests`

src/toxcore/dht_new/server.rs
kurnevsky: It means that we should get it only after matching on packet kind but not before. So we can try to move the client inside the server and get exact client by pk from packet in each handle_* method if we need it. Not sure how exact it will look like though, but perhaps it worth to try.

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,525 @@
...
+        debug!("received packet {:?}", packet);
+        if let DhtPacket::PingRequest(packet) = packet.unwrap().1 {
+            let ping_req_payload = packet.get_payload(&alice.sk).unwrap();
+            assert_eq!(ping_req_payload.id, prq.id);
+        } else {
+            panic!("can not occur");
+        }
+    }
+
+    // handle_ping_resp()
```
kpp: You forgot to mark this function as `#[test]`

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,525 @@
...
+            let pk = Server::get_pk(&nat_ping_res);
+            let mut client = alice.create_client(&addr.clone(), pk, tx.clone());
+            client.ping_id = nat_res.id;
+            assert!(alice.handle_packet((addr, nat_ping_res), client.clone()).wait().is_ok());
+
+            TestResult::passed()
+        }
+    }
+
+    // handle_ping_req()
```
kpp: You forgot to mark this function as `#[test]`

src/toxcore/dht_new/server.rs
NamsooCho: Good Idea.
But I would like to try this idea on next PR.:)

src/toxcore/dht_new/client.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+impl Client {
+    /// create Client object
+    pub fn new(precomputed_key: PrecomputedKey, pk: PublicKey, addr: SocketAddr) -> Client {
+        Client {
+            pk: pk,
+            addr: addr,
+            precomputed_key: precomputed_key,
+            ping_id: 0xFFFFFFFFFFFFFFFF,
```
kpp: why oxfff?

NamsooCho: Because, can't set ping_id of Client to desired value.
So, can't check in test code.

src/toxcore/dht_new/client.rs
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+    // relay_nat_ping_packet()
+    #[test]
+    fn client_relay_nat_ping_packet_test() {
+        let (client, sk, tx, rx) = create_client();
+        let nat_res = NatPingResponse { id: random_u64() };
+        let nat_payload = DhtRequestPayload::NatPingResponse(nat_res);
+        let dht_req = DhtRequest::new(&client.precomputed_key, &client.pk, &client.pk, nat_payload.clone());
+        client.relay_nat_ping_packet(tx, &client.addr, dht_req).wait().unwrap();
```
kpp: Why relay_nat_ping_packet?

NamsooCho: It is used both for nat_ping_request/response.

src/toxcore/dht_new/client.rs
```diff
@@ -0,0 +1,236 @@
...
+        client.send_ping_request(tx).wait().unwrap();
+        let (received_packet, _rx) = rx.into_future().wait().unwrap();
+        if let DhtPacket::PingRequest(packet) = received_packet.unwrap().1 {
+            let ping_req_payload = packet.get_payload(&sk).unwrap();
+            assert_eq!(ping_req_payload.id, client.ping_id);
+        } else {
+            unreachable!("can not occur");
+        }
+    }
+}
```
kpp: add new line at EOF

NamsooCho: Added.

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,630 @@
...
+    /// create new client
+    pub fn create_client(&mut self, addr: &SocketAddr, pk: PublicKey) -> Client {
+        let precomputed_key = encrypt_precompute(&pk, &self.sk);
+        Client::new(precomputed_key, pk, addr.clone())
+    }
+    /**
+    Function to handle incoming packets. If there is a response packet,
+    send back it to the peer.
+    */
+    pub fn handle_packet(&mut self, packet: DhtUdpPacket) -> IoFuture<()>
```
kpp: A DhtUdpPacket is a tuple of (packet, addr). Please [destructure](https://rustbyexample.com/flow_control/match/destructuring.html) it right in arg, so there would be no `packet.1` or `packet.0` calls.

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    /**
+    handle received PingRequest packet, then create PingResponse packet
+    and send back it to the peer.
+    */
+    fn handle_ping_req(&mut self, client: Client, request: PingRequest) -> IoFuture<()>
+    {
+        if let Ok(payload) = request.get_payload(&self.sk) {
+            let resp_payload = PingResponsePayload {
```
kpp: Remove `From<PingRequestPayload> for PingResponsePayload` since you don't use it.

src/toxcore/dht_new/client.rs
kpp: But why did you chose such a value?

NamsooCho: As you can see at https://github.com/NamsooCho/tox-1/blob/7293bfb67bfd02087f4463a438c2a64fcf01df1a/src/toxcore/dht_new/server.rs#L217-L223

0 is checked

NamsooCho: https://github.com/NamsooCho/tox-1/blob/7293bfb67bfd02087f4463a438c2a64fcf01df1a/src/toxcore/dht_new/server.rs#L466-L468

NamsooCho: Client is moved to inside of handle_packet().
So, can't set ping_id.

src/toxcore/dht_new/client.rs
kpp: I still do not understand why you used `relay` word here.

src/toxcore/dht_new/client.rs
NamsooCho: https://github.com/NamsooCho/tox-1/blob/7293bfb67bfd02087f4463a438c2a64fcf01df1a/src/toxcore/dht_new/server.rs#L134-L138

src/toxcore/dht_new/client.rs
NamsooCho: if the rpk is not same as this DHT node, it relay to proper DHT node.

NamsooCho: resend?

src/toxcore/dht_new/client.rs
kpp: In server you check that it should not be 0. But you randomize the id before you actually send ping request: https://github.com/tox-rs/tox/pull/25/files#diff-5c3240e241a145d5d67ead09105441e7R113  . So is there any reason why you had to chose 0xffffff as initial value?

src/toxcore/dht_new/client.rs
kpp: send?

NamsooCho: Ok

src/toxcore/dht_new/client.rs
NamsooCho: https://github.com/NamsooCho/tox-1/blob/7293bfb67bfd02087f4463a438c2a64fcf01df1a/src/toxcore/dht_new/server.rs#L442

https://github.com/NamsooCho/tox-1/blob/7293bfb67bfd02087f4463a438c2a64fcf01df1a/src/toxcore/dht_new/server.rs#L461

In previous commit,
"client.ping_id = prs.id" was possible.

Now, client is not exposed to external.

NamsooCho: So I don't know how to write test code for handle_packet(), especially in *Response()

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,630 @@
...
+
+    DHT `PublicKey` and `SecretKey` are supposed to be ephemeral.
+    */
+    pub fn new(tx: Tx) -> Result<Server, Error> {
+        if !crypto_init() {
+            return Err(Error::new(ErrorKind::Other,
+                       "Crypto initialization failed."));
+        }
+
+        let (pk, sk) = gen_keypair();
```
kpp: You have to gen this info via args, because we want to start a node with the given keypair.

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,630 @@
...
+        Client::new(precomputed_key, pk, addr.clone())
+    }
+    /**
+    Function to handle incoming packets. If there is a response packet,
+    send back it to the peer.
+    */
+    pub fn handle_packet(&mut self, packet: DhtUdpPacket) -> IoFuture<()>
+    {
+        let pk = Server::get_pk(&packet.1.clone());
+        let client = self.create_client(&packet.0, pk);
```
kpp: Have a look at https://github.com/tox-rs/tox/pull/33 . You logic is incorrect for BootstrapInfo packet

src/toxcore/dht_new/client.rs
kurnevsky: But you can't just create a new client for every packet here: https://github.com/NamsooCho/tox-1/blob/7293bfb67bfd02087f4463a438c2a64fcf01df1a/src/toxcore/dht_new/server.rs#L137
You have to store somehow client or it's ping_id. The server should remember this in order to be able to handle ping responses correctly. So if server store it you can set it in tests.

NamsooCho: Ah !!!
The Client object should be created at request time.
When receiving request from other peer, just response with same ping_id.
Ping_id is needed when response is checked.
Client object is enough to be kept just one transaction of req/res.
Right?

kurnevsky: > Ping_id is needed when response is checked.
Yes.

> Client object is enough to be kept just one transaction of req/res.
Server sends pings to each node from it's kbucket periodically. So it should store somehow ping ids for each node from kbucket. And when server receives ping response it checks if ping id is the same as stored or no. If yes then response is considered as correct. Perhaps we shouldn't store client (do we need this entity at all?) but only ping ids for each node in kbucket.

NamsooCho: From implementing perspective, Client is useful for containing Tx(mpsc), SocketAddr, pk, ping_id,...
Current kbucket contains only SocektAddr, PublicKey.
Maybe It is better to extend kbucket.

kurnevsky: > From implementing perspective, Client is useful for containing Tx(mpsc), SocketAddr, pk, ping_id,...


And why does it contain ping_id? It's useless if you have new client for every packet.

NamsooCho: Got it!!
Thanks for your comments.
I will try to remove Client struct.

kpp: And please, don't init ping_id with 0xffff

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,683 @@
...
+    {
+        match packet {
+            DhtPacket::PingRequest(packet) => {
+                debug!("Received ping request");
+                let client = self.create_client(&addr, packet.pk);
+                self.handle_ping_req(client, packet)
+            },
+            DhtPacket::PingResponse(packet) => {
+                debug!("Received ping response");
+                let client = self.get_client(&packet.pk);
```
kpp: It is not 100% for sure the client will be in cache

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,683 @@
...
+                self.handle_ping_resp(client, packet)
+            },
+            DhtPacket::NodesRequest(packet) => {
+                debug!("Received NodesRequest");
+                let client = self.create_client(&addr, packet.pk);
+                self.handle_nodes_req(client, packet)
+            },
+            DhtPacket::NodesResponse(packet) => {
+                debug!("Received NodesResponse");
+                let client = self.get_client(&packet.pk);
```
kpp: It is not 100% for sure the client will be in cache

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,683 @@
...
+                    });
+                match payload {
+                    Ok(DhtRequestPayload::NatPingRequest(pl)) => {
+                        debug!("Received nat ping request");
+                        let client = self.create_client(&addr, dr.spk);
+                        self.handle_nat_ping_req(client, dr, pl)
+                    },
+                    Ok(DhtRequestPayload::NatPingResponse(pl)) => {
+                        debug!("Received nat ping response");
+                        let client = self.get_client(&dr.spk);
```
kpp: It is not 100% for sure the client will be in cache

src/toxcore/dht_new/server.rs
```diff
@@ -0,0 +1,696 @@
...
+        }
+    }
+
+    /// create new client
+    pub fn create_client(&mut self, addr: &SocketAddr, pk: PublicKey) -> Client {
+        let precomputed_key = encrypt_precompute(&pk, &self.sk);
+        Client::new(precomputed_key, pk, addr.clone(), self.tx.clone())
+    }
+    /// get client from cache
+    pub fn get_client(&self, pk: &PublicKey) -> Option<Client> {
```
kpp: `create_client` const you almost zero CPU. Please, don't precache clients for now.

NamsooCho: send PingRequest to peer A with ping_id = 0x01.
send PingRequest to peer B with ping_id = 0x02.
received PingRespinse from peer B -> handle_packet() need client.

src/toxcore/dht_new/client.rs
```diff
@@ -0,0 +1,258 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Tox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+
+/*!
+Hold peer info. connected.
```
kpp: wtf

PR #36, opened 2018-03-01
Remove old dht v1

kpp:
The old dht code was not actually removed. Instead it was moved to `src/toxcore/dht_old`.


kurnevsky:
Maybe we can remove it completely but leave just branch or tag with old dht?

kpp:
> Maybe we can remove it completely

There are some code we still "need": state-format, kbucket tests.

PR #40, opened 2018-03-02
refactor(binary_io): remove it in favor of binary_io_new

kpp:
Remove it in favor of binary_io_new.
Use new binary_io in toxid and in crypto_core_tests.


kurnevsky:
It seems we can rename binary_io_new to binary_io already?

kpp:
> It seems we can rename binary_io_new to binary_io already?

Yes. The next step is to rename it.

PR #50, opened 2018-03-11
Implement onion payloads

kurnevsky:


src/toxcore/onion/packet.rs
```diff
@@ -1019,4 +1560,143 @@ mod tests {
         };
         assert!(invalid_onion_return.get_payload(&symmetric_key).is_err());
     }
+
+    #[test]
+    fn onion_request_0_payload_encrypt_decrypt() {
+        let (alice_pk, alice_sk) = gen_keypair();
+        let (bob_pk, _bob_sk) = gen_keypair();
+        let (_eve_pk, eve_sk) = gen_keypair();
```
kpp: It would be better to create a standalone test to check whether Eve is possible to decode your packets instead of testing this case in each test.

kurnevsky: No, we have different get_payload methods for each packet. So we have to check it in each test to cover them all.
It's possible to split tests for valid/invalid decryption though. But it will only lead to more lines of code - can't see any benefits.

kurnevsky: Then why we check all the packets in one test here https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/codec.rs#L156 ? Shouldn't we split it to separate tests as well?

My point is that if we check decryption with eve's key in separate test it will almoust double the code size because we will create identical situations:
```rust
    #[test]
    fn onion_request_0_payload_encrypt_decrypt() {
        let (alice_pk, alice_sk) = gen_keypair();
        let (bob_pk, _bob_sk) = gen_keypair();
        let shared_secret = encrypt_precompute(&bob_pk, &alice_sk);
        let payload = OnionRequest0Payload {
            ip_port: IpPort {
                ip_addr: "5.6.7.8".parse().unwrap(),
                port: 12345
            },
            temporary_pk: gen_keypair().0,
            inner: vec![42, 123]
        };
        // encode payload with shared secret
        let onion_packet = OnionRequest0::new(&shared_secret, &alice_pk, payload.clone());
        // decode payload with bob's secret key
        let decoded_payload = onion_packet.get_payload(&shared_secret).unwrap();
        // payloads should be equal
        assert_eq!(decoded_payload, payload);
    }

    #[test]
    fn onion_request_0_payload_encrypt_decrypt_invalid_key() {
        let (alice_pk, alice_sk) = gen_keypair();
        let (bob_pk, _bob_sk) = gen_keypair();
        let (_eve_pk, eve_sk) = gen_keypair();
        let shared_secret = encrypt_precompute(&bob_pk, &alice_sk);
        let payload = OnionRequest0Payload {
            ip_port: IpPort {
                ip_addr: "5.6.7.8".parse().unwrap(),
                port: 12345
            },
            temporary_pk: gen_keypair().0,
            inner: vec![42, 123]
        };
        // encode payload with shared secret
        let onion_packet = OnionRequest0::new(&shared_secret, &alice_pk, payload.clone());
        // try to decode payload with eve's secret key
        let eve_shared_secret = encrypt_precompute(&bob_pk, &eve_sk);
        let decoded_payload = onion_packet.get_payload(&eve_shared_secret);
        assert!(decoded_payload.is_err());
    }
```

So you sure that it's worth it?

kpp: https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/mod.rs#L267..L285

kurnevsky: https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/server/server.rs#L414-L494

src/toxcore/onion/packet.rs
```diff
@@ -255,6 +256,95 @@ impl ToBytes for OnionRequest0 {
...
+
+Inner payload should be sent to the next node with address from `ip_port` field.
+
+Serialized form:
+
+Length   | Content
+-------- | ------
+`19`     | `IpPort` of the next node
+`32`     | Temporary `PublicKey`
+variable | Payload
```
kpp: Add comment: `where payload is....` here and below

PR #53, opened 2018-03-13
Codec test

NamsooCho:
Add test case for codec.rs
Remove duplicated test case of packet.rs

### DIFF
--- a/src/toxcore/dht/packet.rs
+++ b/src/toxcore/dht/packet.rs
@@ -1120,10 +1120,6 @@ mod tests {
     use std::net::SocketAddr;
     use quickcheck::{Arbitrary, Gen, quickcheck};
 
-    const ONION_RETURN_1_PAYLOAD_SIZE: usize = ONION_RETURN_1_SIZE - NONCEBYTES;
-    const ONION_RETURN_2_PAYLOAD_SIZE: usize = ONION_RETURN_2_SIZE - NONCEBYTES;
-    const ONION_RETURN_3_PAYLOAD_SIZE: usize = ONION_RETURN_3_SIZE - NONCEBYTES;
-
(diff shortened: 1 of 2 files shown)


kpp:
Add test for: https://coveralls.io/builds/15940383/source?filename=toxcore%2Fdht%2Fcodec.rs#L64

kpp:
This one is missing too https://coveralls.io/builds/15940383/source?filename=toxcore%2Fdht%2Fpacket.rs#L340

src/toxcore/dht/packet.rs
```diff
@@ -1602,6 +1447,16 @@ mod tests {
                 };
                 let decoded_payload = invalid_packet.get_payload(&bob_sk);
                 assert!(decoded_payload.is_err());
+                // Try short incomplete for *Responses
+                let invalid_payload = [0x01];
```
kurnevsky: Why do you need to distinguish requests and responses? Maybe empty payload will fit for both?

kurnevsky: Anyway this part won't be necessary after migration to nom 4.

NamsooCho: I write that code to fix this problem which is mentioned by Roman 
"This one is missing too https://coveralls.io/builds/15940383/source?filename=toxcore%2Fdht%2Fpacket.rs#L340"
If there is any other good solution, let me know.

kurnevsky: Did you try to use invalid_payload = []; ? Because it seems to me that it's not quite correct to use 0x00 and 0x01 for requests/responses because it covers only ping/pong cases.

NamsooCho: OK!
I will try your recommendation on next PR

PR #54, opened 2018-03-13
Revise DHT docs

NamsooCho:
Revise comments for docs.

- packet.rs
- packed_node.rs
- client.rs
- kbucket.rs
- mod.rs
- server.rs

### DIFF
--- a/src/toxcore/dht/packet.rs
+++ b/src/toxcore/dht/packet.rs
@@ -148,6 +148,8 @@ Length  | Content
 `24`    | Nonce
 `9`     | Payload
 
+where Payload is encrypted [`PingRequestPayload`](./struct.PingRequestPayload.html)
+
 */
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct PingRequest {
@@ -287,6 +289,8 @@ Length  | Content
 `32`    | Public Key
 `24`    | Nonce
 `9`     | Payload
+
+where Payload is encrypted [`PingResponsePayload`](./struct.PingResponsePayload.html)
 
 */
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -428,6 +432,8 @@ Length  | Content
 `24`    | Nonce
 `40`    | Payload
 
+where Payload is encrypted [`NodesRequestPayload`](./struct.NodesRequestPayload.html)
+
 */
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct NodesRequest {
@@ -560,6 +566,8 @@ Length    | Content
 `24`      | Nonce
 `1`       | Number of Response Nodes
 `[47,212]`| Payload
+
+where Payload is encrypted [`NodesResponsePayload`](./struct.NodesResponsePayload.html)
 
 */
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


src/toxcore/dht/packet.rs
```diff
@@ -528,6 +555,16 @@ should respond with `NodesResponse` that contains up to to 4 closest nodes to
...
 https://zetok.github.io/tox-spec/#dht-packet
+
+Length    | Content
+--------- | -------------------------
+`1`       | `0x04`
+`32`      | Public Key
+`24`      | Nonce
+`1`       | Number of Response Nodes
+`[39,204]`| Packed Nodes
+`8`       | Request Id (Ping Id)
```
kurnevsky: Just payload will be better here instead of last 3 rows. Because current version doesn't show that payload is encrypted and has additional MACBYTES bytes.

src/toxcore/dht/packet.rs
```diff
@@ -404,6 +422,15 @@ requested public key. Every 20 seconds DHT node sends `NodesRequest` packet to
...
 
 https://zetok.github.io/tox-spec/#dht-packet
+
+Length  | Content
+------- | -------------------------
+`1`     | `0x02`
+`32`    | Public Key
+`24`    | Nonce
+`32`    | Requested Node Public Key
+`8`     | Request Id (Ping Id)
```
kurnevsky: The same here (as for NodesResponse).

src/toxcore/dht/packet.rs
```diff
@@ -272,6 +281,15 @@ respond with `PingResponse` that contains the same ping id inside it's encrypted
...
 
 https://zetok.github.io/tox-spec/#dht-packet
+
+Length  | Content
+------- | -------------------------
+`1`     | `0x01`
+`32`    | Public Key
+`24`    | Nonce
+`1`     | `0x01`
+`8`     | Request Id (Ping Id) in BigEndian
```
kurnevsky: The same here (as for NodesResponse).

src/toxcore/dht/packet.rs
```diff
@@ -140,6 +140,15 @@ doesn't arrive for 122 seconds the DHT node removes peer from kbucket and marks
...
 
 https://zetok.github.io/tox-spec/#dht-packet
+
+Length  | Content
+------- | -------------------------
+`1`     | `0x00`
+`32`    | Public Key
+`24`    | Nonce
+`1`     | `0x00`
+`8`     | Request Id (Ping Id) in BigEndian
```
kurnevsky: The same here (as for NodesResponse).

src/toxcore/dht/packet.rs
NamsooCho: I got your intention.
I think that both of my docs and your intention have good explanation of packet structure.
So, I would like to hear Roman's opinion.

kpp: I agree with Evgeniy.

src/toxcore/dht/packet.rs
```diff
@@ -140,6 +140,14 @@ doesn't arrive for 122 seconds the DHT node removes peer from kbucket and marks
 it as offline if the peer is known friend.
 
 https://zetok.github.io/tox-spec/#dht-packet
+
+Length  | Content
+------- | -------------------------
+`1`     | `0x00`
+`32`    | Public Key
+`24`    | Nonce
+`9`     | Payload
```
kpp: Encrypted Payload, where Payload is .... link to payload struct

kpp: @NamsooCho have a look at https://github.com/tox-rs/tox/pull/50/files#diff-bd6ee0853157076924e3716122a19209R318

PR #62, opened 2018-03-21
Enable osx builds

kurnevsky:


kpp:
Should `osx` be in `allow_failures`?

kurnevsky:
Perhaps due to spoiled nature of osx it might fail builds :)
Actually, I don't know why it was placed in allow_failures. I you don't see any issues - I can remove it from there.

kpp:
Please remove it. We will debug failures, it case of `spoiled nature of osx` we will move it to allowed failures

PR #71, opened 2018-03-27
feat(dht): lan discovery handler

NamsooCho:
It use broadcasting

### DIFF
--- a/src/toxcore/dht/client.rs
+++ b/src/toxcore/dht/client.rs
@@ -24,13 +24,15 @@ Hold infomation of a peer.
 The object of this struct is one per a peer.
 */
 
-use futures::{Sink, Future};
+use futures::{Sink, Future, stream, Stream};
 use futures::sync::mpsc;
 use tokio_io::IoFuture;
+use get_if_addrs;
+use get_if_addrs::IfAddr;
 
 //use std::collections::VecDeque;
 use std::io::{ErrorKind, Error};
-use std::net::SocketAddr;
+use std::net::{SocketAddr, IpAddr};
 
 use toxcore::crypto_core::*;
 use toxcore::dht::packet::*;
@@ -125,12 +127,53 @@ impl Client {
         let nat_ping_req = DhtPacket::DhtRequest(DhtRequest::new(&self.precomputed_key.clone(), &friend_pk, &self.pk, payload));
         self.send(nat_ping_req)
     }
+    // get broadcast addresses for host's network interfaces
+    fn get_ipv4_broadcast_addrs() -> Vec<IpAddr> {
+        let ifs = get_if_addrs::get_if_addrs().expect("no network interface");
+        ifs.iter().filter_map(|interface|
+            match interface.addr {
+                IfAddr::V4(ref addr) => addr.broadcast,
+                _ => None,
+        })
+        .map(|ipv4|
+            IpAddr::V4(ipv4)
+        ).collect()
+    }
+    // get broadcast addresses for Ipv4 and Ipv6
+    fn get_broadcast_addresses() -> Vec<IpAddr> {
+        let mut res = Vec::new();
+        let ipv4s = Client::get_ipv4_broadcast_addrs();
+
+        res.extend(ipv4s.iter());
+
+        // Ipv4 global broadcast address
+        res.push(
+            ("255.255.255.255").parse().unwrap()
+        );
+        // Ipv6 broadcast address
+        res.push(
+            ("FF02::1").parse().unwrap()
+        );
+        res
+    }
+    /// send LanDiscovery packets to broadcast addresses
+    pub fn send_lan_discovery(&self) -> IoFuture<()> {
+        let ip_addrs = Client::get_broadcast_addresses();
+        let lan_packet = DhtPacket::LanDiscovery(LanDiscovery {
+            pk: self.pk,
+        });
+        let lan_sender = ip_addrs.iter().map(|&addr|
+            self.send_to(SocketAddr::new(addr, 33445), lan_packet.clone())
+        );
+
+        let lan_stream = stream::futures_unordered(lan_sender).then(|_| Ok(()));
+        Box::new(lan_stream.for_each(|()| Ok(())))
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::*;
 
     use std::net::SocketAddr;
     use toxcore::dht::packed_node::*;
@@ -291,5 +334,31 @@ mod tests {
         let (_, size) = dht_payload.to_bytes((&mut buf, 0)).unwrap();
         let (_, nat_ping_req_payload) = NatPingRequest::from_bytes(&buf[..size]).unwrap();
         assert_eq!(nat_ping_req_payload.id, client.ping_id);
+    }
+    // send_lan_discovery()
+    #[test]
+    fn client_send_lan_discovery_test() {
+        let (client, _sk, mut rx) = create_client();
+        client.send_lan_discovery().wait().unwrap();
+
+        let ifs = get_if_addrs::get_if_addrs().expect("no network interface");
+        let broad_vec: Vec<SocketAddr> = ifs.iter().filter_map(|interface| 
+            match interface.addr {
+                IfAddr::V4(ref addr) => addr.broadcast,
+                _ => None,
+            })
+            .map(|ipv4|
+                SocketAddr::new(IpAddr::V4(ipv4), 33445)
+            ).collect();
+        for _i in 0..(broad_vec.len() + 2) { // `+2` are one for 255.255.255.255 and one for FF02::1
+            let (received, rx1) = rx.into_future().wait().unwrap();
+            debug!("received packet {:?}", received.clone().unwrap().1);
+            let (packet, _addr) = received.unwrap();
+            let mut buf = [0; 512];
+            let (_, size) = packet.to_bytes((&mut buf, 0)).unwrap();
+            let (_, lan_discovery) = LanDiscovery::from_bytes(&buf[..size]).unwrap();
+            assert_eq!(lan_discovery.pk, client.pk);
+            rx = rx1;
+        }
     }
 }
--- a/src/toxcore/dht/server.rs
+++ b/src/toxcore/dht/server.rs
@@ -228,6 +228,11 @@ impl Server {
                     },
                 }
             },
+            DhtPacket::LanDiscovery(packet) => {
+                debug!("Received LanDiscovery");
+                let client = self.create_client(&addr, packet.pk);
+                self.handle_lan_discovery(client, packet)
+            },
             ref p => {
                 error!("received packet are not handled {:?}", p);
                 Box::new( future::err(
@@ -237,7 +242,6 @@ impl Server {
             }
         }
     }
-
     /**
     handle received PingRequest packet, then create PingResponse packet
     and send back it to the peer.
@@ -406,6 +410,25 @@ impl Server {
                 Box::new( future::ok(()) )
             }
         }
+    }
+    /**
+    handle received LanDiscovery packet, then create NodesRequest packet
+    and send back it to the peer.
+    */
+    fn handle_lan_discovery(&self, mut client: Client, packet: LanDiscovery) -> IoFuture<()> {
+        let mut state = self.state.write().expect("Failed to obtain ServerState lock");
+        let result = client.send_nodes_request(packet.pk);
+        state.peers_cache.insert(packet.pk, client);
+        result
+    }
+    /**
+    send LanDiscovery packet to all broadcast addresses
+    */
+    pub fn send_lan_discovery(&self) -> IoFuture<()> {
+        // addr is dummy, actual lan discovery packet sending logic exists in client
+        let addr: SocketAddr = "127.0.0.1:33445".parse().unwrap();
+        let client = self.create_client(&addr, self.pk);
+        client.send_lan_discovery()
     }
     /// add PackedNode object to kbucket as a thread-safe manner
     pub fn try_add_to_kbucket(&self, pn: &PackedNode) -> bool {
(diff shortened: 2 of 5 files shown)


src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,114 @@
...
+            port,
+        }
+    }
+    fn get_ipv4_broadcast_addrs(&self) -> Vec<SocketAddr> {
+        let mut res = Vec::new();
+        let ifs = pnet_datalink::interfaces();
+        ifs.iter().for_each(|interface|
+            interface.ips.iter()
+                .filter(|ip| ip.is_ipv4())
+                .map(|ip|
```
kurnevsky: You can use filter_map here to avoid unreachable!

NamsooCho: thanks.
I will try.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,114 @@
...
+                .for_each(|broadcast|
+                    res.push(
+                        SocketAddr::new(IpAddr::V4(broadcast), self.port)
+                    )
+                )
+        );
+        res
+    }
+    /// assemble Lan Discovery DhtPackets
+    pub fn assemble_packets(&self) -> Vec<(DhtPacket, SocketAddr)> {
```
kurnevsky: Why this function returns multiple packets? It seems that packet is the same for each address.

src/toxcore/dht/server.rs
```diff
@@ -407,6 +411,31 @@ impl Server {
...
+        let result = client.send_nodes_request(friend_pk);
+        state.peers_cache.insert(packet.pk, client);
+        result
+    }
+    /**
+    send LanDiscovery packet to all broadcast addresses
+    */
+    pub fn send_lan_discovery(&self) -> IoFuture<()> {
+        // addr is dummy, actual lan discovery packet sending logic exists in client
+        let addr: SocketAddr = "127.0.0.1:33445".parse().unwrap();
```
kurnevsky: Yet another reason why we should move tx and send_* functions from client to the server :)

src/toxcore/dht/lan_discovery.rs
NamsooCho: Because, same packet contents need to be sent to multiple addresses.

src/toxcore/dht/server.rs
NamsooCho: What do you mean?

NamsooCho: Ah
I got your saying.
Well, I think your comments deeply

src/toxcore/dht/lan_discovery.rs
kurnevsky: But why don't create this packet once outside of this function? You can return vector of addresses here only.

src/toxcore/dht/server.rs
kpp: It requires a global refactoring. For now it is OK to create a dummy client, but in a week we have to rewrite server to never meet the code like:

```
// addr is dummy, actual lan discovery packet sending logic exists in client
let addr: SocketAddr = "127.0.0.1:33445".parse().unwrap();
```

kurnevsky: I mean that it's redundantly to store rx in each client because we have only one channel for udp connection. And if we remove tx from client and move send_* functions you won't have to create dummy address here. But it's not necessary for this pr.

src/toxcore/dht/lan_discovery.rs
NamsooCho: I revised code according to your comments.
It is much better, now.
Thanks

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,107 @@
...
+                    )
+                )
+        );
+        res
+    }
+    /// assemble Lan Discovery DhtPackets
+    pub fn assemble_packets(&self) -> Vec<SocketAddr> {
+        let mut res = Vec::new();
+        let ipv4s = self.get_ipv4_broadcast_addrs();
+        ipv4s.iter().for_each(|addr|
```
kpp: https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,105 @@
...
+                .for_each(|broadcast|
+                    res.push(
+                        SocketAddr::new(IpAddr::V4(broadcast), self.port)
+                    )
+                )
+        );
+        res
+    }
+    /// assemble Lan Discovery DhtPackets
+    pub fn assemble_packets(&self) -> Vec<SocketAddr> {
```
kurnevsky: This function has wrong name now. Perhaps it can be unified with get_ipv4_broadcast_addrs?

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/** Structure that is needed to assemble Lan Discovery packet.
+The packet assembled is sent to Broadcast address of LAN.
+The broadcast addresses are interface broadcast and global boradcast(255.255.255.255) address for Ipv4,
+and multicast(FF02::1) address for Ipv6.
+*/
+#[derive(Debug)]
+pub struct LanDiscoveryAssembler {
+    /// PublicKey of DHT server.
+    pk: PublicKey,
```
kurnevsky: There is no need in pk anymore.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,105 @@
...
+use pnet_datalink;
+use ipnetwork::*;
+
+/** Structure that is needed to assemble Lan Discovery packet.
+The packet assembled is sent to Broadcast address of LAN.
+The broadcast addresses are interface broadcast and global boradcast(255.255.255.255) address for Ipv4,
+and multicast(FF02::1) address for Ipv6.
+*/
+#[derive(Debug)]
+pub struct LanDiscoveryAssembler {
```
kurnevsky: Name of this struct does not reflect the essence anymore (as well as docs). Do you think it's still needed at all?

src/toxcore/dht/client.rs
```diff
@@ -125,12 +126,27 @@ impl Client {
...
     }
+    /// send LanDiscovery packets to broadcast addresses
+    pub fn send_lan_discovery(&self) -> IoFuture<()> {
+        // TODO: port should be changeable
+        let lan = LanDiscoveryAssembler::new(self.pk, 33445);
+        let packets = lan.get_broadcast_addresses();
+        let lan_packet = DhtPacket::LanDiscovery(LanDiscovery {
+            pk: self.pk,
+        });
+        let lan_sender = packets.iter()
```
kpp: `.iter().map(|x|` on the same line

src/toxcore/dht/client.rs
```diff
@@ -125,12 +126,27 @@ impl Client {
...
+            .map(|&addr|
+            self.send_to(addr, lan_packet.clone())
```
kpp: indent

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,105 @@
...
+        let ifs = pnet_datalink::interfaces();
+        ifs.iter().for_each(|interface|
+            interface.ips.iter()
+                .filter_map(|ip|
+                    match *ip {
+                        IpNetwork::V4(ip) => Some(ip.broadcast()),
+                        _ => None,
+                    }
+                )
+                .for_each(|broadcast|
```
kpp: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map

src/toxcore/dht/client.rs
```diff
@@ -125,12 +127,54 @@ impl Client {
         let nat_ping_req = DhtPacket::DhtRequest(DhtRequest::new(&self.precomputed_key.clone(), &friend_pk, &self.pk, payload));
         self.send(nat_ping_req)
     }
+    // get broadcast addresses for host's network interfaces
+    fn get_ipv4_broadcast_addrs(&self) -> Vec<SocketAddr> {
```
kurnevsky: No need to pass self.

src/toxcore/dht/client.rs
```diff
@@ -125,12 +127,65 @@ impl Client {
...
+                        _ => None,
+                    }
+                })
+                .for_each(|oct| {
+                    let ipaddr_str = format!("{}.{}.{}.{}", oct[0], oct[1], oct[2], oct[3]);
+                    let broad = IPAddress::parse(ipaddr_str).unwrap().broadcast().to_string();
+                    let broad_str: Vec<&str> = broad.split('/').collect();
+                    let broad_str= broad_str[0].to_owned() + ":33445";
+                    let sock: SocketAddr = broad_str.parse().unwrap();
+                    res.push(sock);
```
kpp: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map

Mutable variable `res` is a bad design.

src/toxcore/dht/client.rs
```diff
@@ -125,12 +127,65 @@ impl Client {
...
+    // get broadcast addresses for host's network interfaces
+    fn get_ipv4_broadcast_addrs() -> Vec<SocketAddr> {
+        let ifs = pnet_datalink::interfaces();
+        let mut res = Vec::new();
+        ifs.iter().for_each(|interface| 
+            interface.ips
+                .iter()
+                .filter_map(|ip| {
+                    match ip.ip() {
+                        IpAddr::V4(ip) => Some(ip.octets()),
```
kpp: Do you serialize octests to parse them as ip addr again? Oo

src/toxcore/dht/client.rs
NamsooCho: Well, now I focus at CI build pass.

src/toxcore/dht/client.rs
NamsooCho: To get broadcast address.
I will try better way.

NamsooCho: I am focusing CI build pass

kurnevsky: Why did you get rid of previous code where you just called broadcast method?

NamsooCho: Because, I move from ipnetwork crate to ipaddress crate

NamsooCho: Changing crate is for CI build pass

NamsooCho: The source pushed is not for merge, just test if it pass CI build

kurnevsky: ok, but ipnetwork doesn't depend on external libraries so it shouldn't cause any problems. Problem may be in pnet_datalink though.

src/toxcore/dht/client.rs
```diff
@@ -125,12 +127,53 @@ impl Client {
...
     }
+    // get broadcast addresses for host's network interfaces
+    fn get_ipv4_broadcast_addrs() -> Vec<SocketAddr> {
+        let ifs = get_if_addrs::get_if_addrs().expect("no network interface");
+        ifs.iter().filter_map(|interface| 
+            match interface.addr {
+                IfAddr::V4(ref addr) => addr.broadcast,
+                _ => None,
+            })
+            .map(|ipv4|
```
kpp: indent 4 spaces to the left

src/toxcore/dht/client.rs
```diff
@@ -125,12 +127,53 @@ impl Client {
...
+    // get broadcast addresses for Ipv4 and Ipv6
+    fn get_broadcast_addresses() -> Vec<SocketAddr> {
+        let mut res = Vec::new();
+        let ipv4s = Client::get_ipv4_broadcast_addrs();
+
+        res.extend(ipv4s.iter());
+
+        // Ipv4 global broadcast address
+        res.push(
+            SocketAddr::new(("255.255.255.255").parse().unwrap(), 33445)
```
kpp: Why 33445? You'd better return only ip addrs without ip, set ip only when you actually send it

src/toxcore/dht/client.rs
```diff
@@ -125,12 +127,53 @@ impl Client {
...
+        );
+        // Ipv6 broadcast address
+        res.push(
+            SocketAddr::new(("FF02::1").parse().unwrap(), 33445)
+        );
+        res
+    }
+    /// send LanDiscovery packets to broadcast addresses
+    pub fn send_lan_discovery(&self) -> IoFuture<()> {
+        let packets = Client::get_broadcast_addresses();
```
kpp: really `let packets`?

src/toxcore/dht/client.rs
```diff
@@ -292,4 +335,30 @@ mod tests {
...
+        let ifs = get_if_addrs::get_if_addrs().expect("no network interface");
+        let broad_vec: Vec<SocketAddr> = ifs.iter().filter_map(|interface| 
+            match interface.addr {
+                IfAddr::V4(ref addr) => addr.broadcast,
+                _ => None,
+            })
+            .map(|ipv4|
+                SocketAddr::new(IpAddr::V4(ipv4), 33445)
+            ).collect();
+        for _i in 0..(broad_vec.len() + 2) {
```
kpp: add comment why `+2`: `// one for 255.255.255.255 and one for FF02::1`

src/toxcore/dht/server.rs
```diff
@@ -407,6 +411,31 @@ impl Server {
             }
         }
     }
+    /**
+    handle received LanDiscovery packet, then create NodesRequest packet
+    and send back it to the peer.
+    */
+    fn handle_lan_discovery(&self, mut client: Client, packet: LanDiscovery) -> IoFuture<()> {
+        // TODO: get real friend pk
```
kpp: Fix it

PR #73, opened 2018-03-31
feat(dht): remove timed out clients

NamsooCho:
Remove clients which is timed-out for 182 seconds

### DIFF
--- a/examples/dht_node.rs
+++ b/examples/dht_node.rs
@@ -135,6 +135,7 @@ fn main() {
     let server = add_lan_sender(server, &server_obj, local);
     let server = add_nodes_sender(server, &server_obj);
     let server = add_onion_key_refresher(server, &server_obj);
+    let server = add_timedout_remover(server, &server_obj);
 
     let server = server
         .map(|_| ())
@@ -225,6 +226,26 @@ fn add_lan_sender(base_selector: IoFuture<()>, server_obj: &Server, local: Socke
         .map_err(|_err| Error::new(ErrorKind::Other, "LanDiscovery timer error"));
     
     Box::new(base_selector.select(Box::new(lan_sender))
+        .map(|_| ())
+        .map_err(move |(err, _select_next)| {
+            error!("Processing ended with error: {:?}", err);
+            err
+        }))
+}
+fn add_timedout_remover(base_selector: IoFuture<()>, server_obj: &Server) -> IoFuture<()> {
+    // 1 seconds for timed-out clients remover
+    const PING_TIMEDOUT_DURATION: u64 = 182;
+    let timeout_dur = Duration::from_secs(PING_TIMEDOUT_DURATION);
+
+    let timeout_wakeups = Interval::new(Duration::from_secs(1));
+    let server_obj_c = server_obj.clone();
+    let timeout_remover = timeout_wakeups.for_each(move |()| {
+            println!("timeout_wakeups");
+            server_obj_c.remove_timedout_clients(timeout_dur)
+        })
+        .map_err(|_err| Error::new(ErrorKind::Other, "Timedout clients remover timer error"));
+    
+    Box::new(base_selector.select(Box::new(timeout_remover))
         .map(|_| ())
         .map_err(move |(err, _select_next)| {
             error!("Processing ended with error: {:?}", err);
--- a/src/toxcore/dht/client.rs
+++ b/src/toxcore/dht/client.rs
@@ -33,6 +33,7 @@ use get_if_addrs::IfAddr;
 //use std::collections::VecDeque;
 use std::io::{ErrorKind, Error};
 use std::net::{SocketAddr, IpAddr};
+use std::time::Instant;
 
 use toxcore::crypto_core::*;
 use toxcore::dht::packet::*;
@@ -53,6 +54,8 @@ pub struct Client {
     pub precomputed_key: PrecomputedKey,
     /// shaed mpsc tx part
     pub tx: Tx,
+    /// last received ping-response time
+    pub last_resp_time: Instant
 }
 
 impl Client {
@@ -64,6 +67,7 @@ impl Client {
             precomputed_key,
             ping_id: 0,
             tx,
+            last_resp_time: Instant::now(),
         }
     }
     /// actual send method
--- a/src/toxcore/dht/server.rs
+++ b/src/toxcore/dht/server.rs
@@ -36,6 +36,7 @@ use std::io::{ErrorKind, Error};
 use std::net::SocketAddr;
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 
 use toxcore::crypto_core::*;
 use toxcore::dht::packet::*;
@@ -133,6 +134,20 @@ impl Server {
         else {
             None
         }
+    }
+    /// remove timed-out clients, also remove node from kbucket
+    pub fn remove_timedout_clients(&self, timeout: Duration) -> IoFuture<()> {
+        let mut state = self.state.write();
+        let timeout_peers = state.peers_cache.iter()
+            .filter(|&(_pk, ref client)| client.last_resp_time.elapsed() > timeout)
+            .map(|(pk, _client)| *pk)
+            .collect::<Vec<_>>();
+
+        timeout_peers.into_iter().for_each(|pk| state.kbucket.remove(&pk));
+
+        state.peers_cache.retain(|&_pk, ref client|
+            client.last_resp_time.elapsed() <= timeout);
+        Box::new(future::ok(()))
     }
     /// send PingRequest to all peers in kbucket
     pub fn send_pings(&self) -> IoFuture<()> {
@@ -282,7 +297,7 @@ impl Server {
     */
     fn handle_ping_resp(&self, packet: PingResponse) -> IoFuture<()>
     {
-        let state = self.state.read();
+        let mut state = self.state.write();
         let payload = packet.get_payload(&self.sk);
         let payload = match payload {
             Err(e) => {
@@ -301,7 +316,7 @@ impl Server {
             )))
         }
         let client = self.get_client(&packet.pk, &state.peers_cache);
-        let client = match client {
+        let mut client = match client {
             None => {
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
@@ -312,6 +327,8 @@ impl Server {
         };
 
         if client.ping_id == payload.id {
+            client.last_resp_time = Instant::now();
+            state.peers_cache.insert(packet.pk, client);
             Box::new( future::ok(()) )
         }
         else {
@@ -381,7 +398,7 @@ impl Server {
             None => {
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
-                        "get_client() fail upon NodesResponse"
+                        "get_client() failed in handle_nodes_resp()"
                 )))
             },
             Some(client) => client,
@@ -433,7 +450,7 @@ impl Server {
             None => {
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
-                        "get_client() fail upon NatPingResponse"
+                        "get_client() failed in handle_nat_ping_resp()"
                 )))
             },
             Some(client) => client,
@@ -1676,5 +1693,43 @@ mod tests {
         let (_, size) = packet.to_bytes((&mut buf, 0)).unwrap();
         let (_, lan_discovery) = LanDiscovery::from_bytes(&buf[..size]).unwrap();
         assert_eq!(alice.pk, lan_discovery.pk);
+    }
+    // remove_timedout_clients(), case of client removed
+    #[test]
+    fn server_remove_timedout_clients_removed_test() {
+        let (alice, _precomp, bob_pk, _bob_sk, _rx, _addr) = create_node();
+        let node = PackedNode::new(false, SocketAddr::V4("127.0.0.1:12345".parse().unwrap()), &bob_pk);
+
+        alice.try_add_to_kbucket(&node);
+
+        alice.send_pings().wait().unwrap();
+
+        let dur = Duration::from_secs(0);
+        alice.remove_timedout_clients(dur).wait().unwrap();
+        let state = alice.state.read();
+        // after client be removed
+        assert!(alice.get_client(&bob_pk, &state.peers_cache).is_none());
+        // peer should be removed from kbucket
+        let state = alice.state.read();
+        assert!(!state.kbucket.contains(&bob_pk));
+    }
+    // remove_timedout_clients(), case of client remained
+    #[test]
+    fn server_remove_timedout_clients_remained_test() {
+        let (alice, _precomp, bob_pk, _bob_sk, _rx, _addr) = create_node();
+        let node = PackedNode::new(false, SocketAddr::V4("127.0.0.1:12345".parse().unwrap()), &bob_pk);
+
+        alice.try_add_to_kbucket(&node);
+
+        alice.send_pings().wait().unwrap();
+
+        let dur = Duration::from_secs(1);
+        alice.remove_timedout_clients(dur).wait().unwrap();
+        let state = alice.state.read();
+        // client should be remained
+        assert!(!alice.get_client(&bob_pk, &state.peers_cache).is_none());
+        // peer should be remained in kbucket
+        let state = alice.state.read();
+        assert!(state.kbucket.contains(&bob_pk));
     }
 }


examples/dht_node.rs
```diff
@@ -152,6 +152,15 @@ fn main() {
         })
         .map_err(|_err| Error::new(ErrorKind::Other, "NatPing timer error"));
 
+    // 3 seconds for NatPingRequest
```
kpp: 3 for nat?

src/toxcore/dht/server.rs
```diff
@@ -338,6 +356,7 @@ impl Server {
                 for node in &payload.nodes {
                     state.kbucket.try_add(node);
                 }
+                client.last_resp_time = Instant::now();
```
kpp: Update only on ping response by design.

NamsooCho: https://zetok.github.io/tox-spec/#dht-operation

Nodes are removed after 122 seconds of no response. Nodes are added to the lists after a valid ping response or send node packet is received from them. If the node is already present in the list it is updated if the IP address changed.

I think above statements is for both PingResponse and NodesResponse

examples/dht_node.rs
NamsooCho: https://zetok.github.io/tox-spec/#hole-punching

NAT ping requests are sent every 3 seconds in toxcore. If no response is received for 6 seconds, the hole-punching will stop.

src/toxcore/dht/server.rs
kpp: Don't think, find corresponding code in C tox impl and compare it to docs.

examples/dht_node.rs
kpp: That's right, but the next line is `let remove_wakeup`.

src/toxcore/dht/server.rs
```diff
@@ -982,4 +1004,22 @@ mod tests {
...
+        let (alice, _precomp, _bob_pk, _bob_sk, _rx, _addr) = create_node();
+        let node = PackedNode::new(false, SocketAddr::V4("127.0.0.1:12345".parse().unwrap()), &alice.pk);
+
+        alice.try_add_to_kbucket(&node);
+
+        alice.send_pings().wait().unwrap();
+        // timeout is 122 seconds
+        let dur = time::Duration::from_secs(123);
+
+        thread::sleep(dur);
```
kurnevsky: I don't think we want to have tests that will run more than 2 minutes. It's better to change client timestamp in the test.

src/toxcore/dht/server.rs
```diff
@@ -123,14 +128,24 @@ impl Server {
             None
         }
     }
+    /// remove timed-out clients
+    pub fn remove_timedout_client(&self) -> IoFuture<()> {
+        let mut state = self.state.write().expect("Failed to obtain ServerState lock");
+        state.peers_cache.retain(|&_pk, ref client|
+            client.last_resp_time.elapsed() <= Duration::from_secs(122));
```
kurnevsky: Make a const please instead of magic number.

src/toxcore/dht/client.rs
```diff
@@ -53,6 +54,8 @@ pub struct Client {
     pub precomputed_key: PrecomputedKey,
     /// shaed mpsc tx part
     pub tx: Tx,
+    /// last received ping-response time
+    pub last_resp_time: Instant
```
kpp: @kurnevsky should we use SystemTime or Instant here?

src/toxcore/dht/server.rs
```diff
@@ -123,14 +128,24 @@ impl Server {
             None
         }
     }
+    /// remove timed-out clients
+    pub fn remove_timedout_client(&self) -> IoFuture<()> {
```
kpp: client**s**

src/toxcore/dht/server.rs
kpp: Or better make it argument for function

src/toxcore/dht/server.rs
```diff
@@ -123,14 +128,24 @@ impl Server {
...
+        let mut state = self.state.write().expect("Failed to obtain ServerState lock");
+        state.peers_cache.retain(|&_pk, ref client|
+            client.last_resp_time.elapsed() <= Duration::from_secs(122));
+        Box::new(future::ok(()))
+    }
     /// send PingRequest to all peer in kbucket    
     pub fn send_pings(&self) -> IoFuture<()> {
-        let mut state = self.state.write().expect("Failed to obtain ServerState lock");
```
kpp: was it a fix for Lan Discovery? Why do you change this code?

src/toxcore/dht/server.rs
```diff
@@ -276,6 +296,7 @@ impl Server {
                 };
                 let mut state = self.state.write().expect("Failed to obtain ServerState lock");
                 state.kbucket.try_add(&packed_node);
+                client.last_resp_time = Instant::now();
```
kpp: How do you update the client? Since you get only `mut Client` not `&mut Client`, is is not updated in you `peer_cache`.

src/toxcore/dht/server.rs
```diff
@@ -982,4 +1004,22 @@ mod tests {
         let (_, lan_discovery) = LanDiscovery::from_bytes(&buf[..size]).unwrap();
         assert_eq!(alice.pk, lan_discovery.pk);
     }
+    // remove_timedout_client()
+    #[test]
+    fn server_remove_timedout_client_test() {
```
kpp: Add one more test that client with ping response (that was 1 sec ago) is not removed from peer cache.

src/toxcore/dht/client.rs
kurnevsky: Another solution - if we pass timeout as argument (as you proposed) we can set it to 0 in tests to expire all clients.

src/toxcore/dht/server.rs
NamsooCho: For remove timed-out clients, create_client() has been changed.
The change is from (just create client) to (check if it exists in peers_cache).
If client already exists, it means that PingRequest() is sent and client had been created and inserted to peers_cache.
Because ping period is 60 seconds and time-out period is 122 seconds.
If every PingRequest create new client, then new client replace old client with same PK.
To do this, change happen in create_client() and to avoid deadlock send_pings() logic is changed.

src/toxcore/dht/server.rs
```diff
@@ -139,12 +154,17 @@ impl Server {
 
         Box::new(pings_stream.for_each(|()| Ok(())))
     }
+    // get random node from kbucket
```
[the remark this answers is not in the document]
NamsooCho: It is for avoiding deadlock.
It is separating logic into 2 function.
If it should be another PR, then I will try other method.

src/toxcore/dht/server.rs
kpp: OK

src/toxcore/dht/server.rs
```diff
@@ -983,4 +1005,34 @@ mod tests {
...
+    #[test]
+    fn server_remove_timedout_clients_removed_test() {
+        let (alice, _precomp, bob_pk, _bob_sk, _rx, _addr) = create_node();
+        let node = PackedNode::new(false, SocketAddr::V4("127.0.0.1:12345".parse().unwrap()), &bob_pk);
+
+        alice.try_add_to_kbucket(&node);
+
+        alice.send_pings().wait().unwrap();
+        let dur = time::Duration::from_secs(2);
+        thread::sleep(dur);
```
kurnevsky: Maybe we shouldn't freeze test at all? Just use alice.remove_timedout_clients(0).

src/toxcore/dht/server.rs
```diff
@@ -124,14 +129,24 @@ impl Server {
             None
         }
     }
+    /// remove timed-out clients
+    pub fn remove_timedout_clients(&self, timeout: u64) -> IoFuture<()> {
```
kpp: timeout must be `Duration`

src/toxcore/dht/server.rs
kpp: Namsoo, try it. But it is optional since 2 secs != 122 =)

src/toxcore/dht/server.rs
```diff
@@ -124,14 +129,24 @@ impl Server {
...
+    /// remove timed-out clients
+    pub fn remove_timedout_clients(&self, timeout: Duration) -> IoFuture<()> {
+        let mut state = self.state.write();
+        state.peers_cache.retain(|&_pk, ref client|
+            client.last_resp_time.elapsed() <= timeout);
+        Box::new(future::ok(()))
+    }
     /// send PingRequest to all peer in kbucket    
     pub fn send_pings(&self) -> IoFuture<()> {
-        let mut state = self.state.write();
```
kpp: Last PR you wrote this code: https://github.com/tox-rs/tox/commit/992eceb89db3b7b995cb13e1210a962c83825ae6#diff-38c05bf285637be9e25083096c19d4e0L128

Please revert it, don't make these changes in the current PR. 1 write lock per method will be fine

src/toxcore/dht/server.rs
```diff
@@ -142,10 +157,14 @@ impl Server {
     }
     /// send NodesRequest to random peer at every 20 seconds
     pub fn send_nodes_req(&self, friend_pk: PublicKey) -> IoFuture<()> {
-        let mut state = self.state.write();
```
kurnevsky: And the same thing here as @kpp wrote - 1 write lock per method will be fine because random_node almost always is Some.

src/toxcore/dht/server.rs
NamsooCho: For remove timed-out clients, create_client() has been changed.
The change is from (just create client) to (check if it exists in peers_cache).
If client already exists, it means that PingRequest() is sent and client had been created and inserted to peers_cache.
Because ping period is 60 seconds and time-out period is 122 seconds.
If every PingRequest create new client, then new client replace old client with same PK.
To do this, change happen in create_client() and to avoid deadlock send_pings() logic is changed.

src/toxcore/dht/server.rs
NamsooCho: For remove timed-out clients, create_client() has been changed.
The change is from (just create client) to (check if it exists in peers_cache).
If client already exists, it means that PingRequest() is sent and client had been created and inserted to peers_cache.
Because ping period is 60 seconds and time-out period is 122 seconds.
If every PingRequest create new client, then new client replace old client with same PK.
To do this, change happen in create_client() and to avoid deadlock send_pings() logic is changed.

src/toxcore/dht/server.rs
kpp: there will be no deadlocks since `write` method waits untill the resourses are released and are free to be used

kpp: Please try to implement it with 1 write lock since calling to `.write()` in loop is a very heavy operation

src/toxcore/dht/server.rs
```diff
@@ -109,28 +110,41 @@ impl Server {
...
+        state.peers_cache.retain(|&_pk, ref client|
+            client.last_resp_time.elapsed() <= timeout);
+        Box::new(future::ok(()))
+    }
     /// send PingRequest to all peer in kbucket    
     pub fn send_pings(&self) -> IoFuture<()> {
-        let mut state = self.state.write();
-        let kbucket_c = state.kbucket.iter();
+        let kbucket_c = {
+            let state = self.state.read();
```
kpp: No need to call state.read since there is a write lock below 3 lines

src/toxcore/dht/server.rs
```diff
@@ -142,9 +156,13 @@ impl Server {
     }
     /// send NodesRequest to random peer at every 20 seconds
     pub fn send_nodes_req(&self, friend_pk: PublicKey) -> IoFuture<()> {
+        let random_node = {
+            let state = self.state.read();
```
kpp: No need to call state.read since there is a write lock below 3 lines

src/toxcore/dht/server.rs
```diff
@@ -170,26 +188,33 @@ impl Server {
...
                     Box::new(future::ok(()))
                 }
             },
             DhtPacket::NodesRequest(packet) => {
                 debug!("Received NodesRequest");
-                let client = self.create_client(&addr, packet.pk);
-                self.handle_nodes_req(client, packet)
+                let state = self.state.read();
+                let client = self.create_client(&addr, packet.pk, &state.peers_cache);
+                self.handle_nodes_req(client, packet, &state.kbucket)
```
kpp: Why don't you
1. call self.handle_nodes_req(packet)
2. inside handle_nodes_req you 1) take read lock 2) create client. 3) continue processing?

src/toxcore/dht/server.rs
```diff
@@ -358,16 +383,15 @@ impl Server {
     /**
     handle received NatPingRequest packet, respond with NatPingResponse
     */
-    pub fn handle_nat_ping_req(&self, client: Client, request: DhtRequest, payload: NatPingRequest) -> IoFuture<()> {
+    pub fn handle_nat_ping_req(&self, client: Client, request: DhtRequest, payload: NatPingRequest, kbucket: &Kbucket) -> IoFuture<()> {
```
kpp: make all handle_* functions private expept for handle_packet

src/toxcore/dht/server.rs
```diff
@@ -839,38 +878,39 @@ mod tests {
         let nat_res = NatPingResponse { id: random_u64() };
         let nat_payload = DhtRequestPayload::NatPingResponse(nat_res);
         let dht_req = DhtRequest::new(&precomp, &bob_pk, &bob_pk, nat_payload.clone());
-        let client = alice.create_client(&addr, bob_pk);
-        assert!(alice.handle_nat_ping_resp(client, dht_req, nat_res).wait().is_ok());
+        let mut state = alice.state.write();
+        let client = alice.create_client(&addr, bob_pk, &state.peers_cache);
+        assert!(alice.handle_nat_ping_resp(client, dht_req, nat_res, &state.kbucket).wait().is_ok());
```
kpp: You should test it by passing special packet, not by calling `handle_nat_ping_resp` directly. See https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/server/server.rs#L430

NamsooCho: Test exists which is like above link.
https://github.com/tox-rs/tox/blob/09f1278d6f957bf51bd0ba9628eac0252739e971/src/toxcore/dht/server.rs#L630-L643

src/toxcore/dht/server.rs
```diff
@@ -170,26 +179,29 @@ impl Server {
...
                 debug!("Received ping request");
-                let client = self.create_client(&addr, packet.pk);
-                self.handle_ping_req(client, packet)
+                self.handle_ping_req(packet, addr)
             },
             DhtPacket::PingResponse(packet) => {
                 debug!("Received ping response");
-                if let Some(client) = self.get_client(&packet.pk) {
-                    self.handle_ping_resp(client, packet)
+                let mut state = self.state.write();
```
kpp: You must call `handle_*(packet, addr)` here, while you take locks inside handle_* function.

src/toxcore/dht/server.rs
```diff
@@ -170,26 +179,29 @@ impl Server {
...
-                let client = self.create_client(&addr, packet.pk);
-                self.handle_nodes_req(client, packet)
+                self.handle_nodes_req(packet, addr)
             },
             DhtPacket::NodesResponse(packet) => {
                 debug!("Received NodesResponse");
-                if let Some(client) = self.get_client(&packet.pk) {
-                    self.handle_nodes_resp(client, packet)
+                let mut state = self.state.write();
+                if let Some(client) = self.get_client(&packet.pk, &state.peers_cache) {
```
kpp: You must call `handle_*(packet, addr)` here, while you take locks inside handle_* function.

src/toxcore/dht/server.rs
```diff
@@ -205,29 +217,29 @@ impl Server {
...
-                        let client = self.create_client(&addr, dr.spk);
-                        self.handle_nat_ping_req(client, dr, pl)
+                        self.handle_nat_ping_req(dr, pl, addr)
                     },
                     Ok(DhtRequestPayload::NatPingResponse(pl)) => {
                         debug!("Received nat ping response");
-                        if let Some(client) = self.get_client(&dr.spk) {
-                            self.handle_nat_ping_resp(client, dr, pl)
+                        let state = self.state.read();
+                        if let Some(client) = self.get_client(&dr.spk, &state.peers_cache) {
```
kpp: You must call `handle_*(packet, addr)` here, while you take locks inside handle_* function.

src/toxcore/dht/server.rs
```diff
@@ -692,28 +723,29 @@ mod tests {
         // handle ping response, request from bob peer
         let prs = PingResponsePayload { id: random_u64() };
         let ping_resp = PingResponse::new(&precomp, &bob_pk, prs);
-        let mut client = alice.create_client(&addr, bob_pk);
+        let mut state = alice.state.write();
+        let mut client = alice.create_client(&addr, bob_pk, &state.peers_cache);
         client.ping_id = prs.id;
-        assert!(alice.handle_ping_resp(client, ping_resp).wait().is_ok());
+        assert!(alice.handle_ping_resp(client, ping_resp, &mut state.kbucket).wait().is_ok());
```
kpp: You should test it by passing special packet and calling `handle_packet`, not by calling handle_YOUR_PACKET directly. See https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/server/server.rs#L430

src/toxcore/dht/server.rs
```diff
@@ -723,17 +755,15 @@ mod tests {
         // error case, empty kbucket
         let nrq = NodesRequestPayload { pk: bob_pk, id: random_u64() };
         let nodes_req = NodesRequest::new(&precomp, &bob_pk, nrq);
-        let client = alice.create_client(&addr, bob_pk);
-        assert!(alice.handle_nodes_req(client, nodes_req).wait().is_err());
+        assert!(alice.handle_nodes_req(nodes_req, addr).wait().is_err());
```
kpp: You should test it by passing special packet and calling `handle_packet`, not by calling handle_YOUR_PACKET directly. See https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/server/server.rs#L430

src/toxcore/dht/server.rs
```diff
@@ -758,38 +787,39 @@ mod tests {
...
         // ping_id = 0
         let nrs = NodesResponsePayload { nodes: vec![
             PackedNode::new(false, SocketAddr::V4("127.0.0.1:12345".parse().unwrap()), &gen_keypair().0)
         ], id: 0 };
         let nodes_resp = NodesResponse::new(&precomp, &bob_pk, nrs.clone());
-        let mut client = alice.create_client(&addr, bob_pk);
+        let mut client = alice.create_client(&addr, bob_pk, &state.peers_cache);
         client.ping_id = 0;
-        assert!(alice.handle_nodes_resp(client.clone(), nodes_resp.clone()).wait().is_err());
+        assert!(alice.handle_nodes_resp(client.clone(), nodes_resp.clone(), &mut state.kbucket).wait().is_err());
```
kpp: You should test it by passing special packet and calling `handle_packet`, not by calling handle_YOUR_PACKET directly. See https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/server/server.rs#L430

src/toxcore/dht/server.rs
```diff
@@ -839,38 +866,39 @@ mod tests {
...
 
-        alice.try_add_to_kbucket(&pn);
+        alice.try_add_to_kbucket(&pn, &mut state.kbucket);
         let nat_res = NatPingResponse { id: random_u64() };
         let nat_payload = DhtRequestPayload::NatPingResponse(nat_res);
         let dht_req = DhtRequest::new(&precomp, &bob_pk, &bob_pk, nat_payload.clone());
-        let client = alice.create_client(&addr, bob_pk);
-        assert!(alice.handle_nat_ping_resp(client, dht_req, nat_res).wait().is_ok());
+        let client = alice.create_client(&addr, bob_pk, &state.peers_cache);
+        assert!(alice.handle_nat_ping_resp(client, dht_req, nat_res, &state.kbucket).wait().is_ok());
```
kpp: You should test it by passing special packet and calling `handle_packet`, not by calling handle_YOUR_PACKET directly. See https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/server/server.rs#L430

src/toxcore/dht/server.rs
```diff
@@ -421,22 +443,28 @@ impl Server {
...
+        let state = self.state.read();
         // addr is dummy, actual lan discovery packet sending logic exists in client
         let addr: SocketAddr = "[::1]:33445".parse().unwrap(); // 33445 is default port for tox
-        let client = self.create_client(&addr, self.pk);
+        let client = self.create_client(&addr, self.pk, &state.peers_cache);
         client.send_lan_discovery_ipv6()
     }
     /// add PackedNode object to kbucket as a thread-safe manner
-    pub fn try_add_to_kbucket(&self, pn: &PackedNode) -> bool {
+    pub fn try_add_to_kbucket(&self, pn: &PackedNode, kbucket: &mut Kbucket) -> bool {
```
kpp: This is method does nothing. It must be removed.

src/toxcore/dht/server.rs
```diff
@@ -109,28 +110,36 @@ impl Server {
...
+        if let Some(client) = peers_cache.get(pk) {
             Some(client.clone())
         }
         else {
             None
         }
     }
-    /// send PingRequest to all peer in kbucket    
+    /// remove timed-out clients
+    pub fn remove_timedout_clients(&self, timeout: Duration) -> IoFuture<()> {
```
kurnevsky: Shouldn't we remove them from kbucket as well?

src/toxcore/dht/server.rs
```diff
@@ -109,28 +110,36 @@ impl Server {
...
+        let mut state = self.state.write();
+        state.peers_cache.retain(|&_pk, ref client|
+            client.last_resp_time.elapsed() <= timeout);
+        Box::new(future::ok(()))
```
kurnevsky: It seems no need to return future here.

src/toxcore/dht/server.rs
```diff
@@ -261,114 +255,141 @@ impl Server {
...
+            )))
+        }
+        let mut client = client.unwrap();
+
+        if client.ping_id == payload.id {
+            let packed_node = PackedNode {
+                saddr: addr,
+                pk: packet.pk,
+            };
+            state.kbucket.try_add(&packed_node);
```
kpp: Why do you call `state.kbucket.try_add` here?

kpp:
Thanks. Looks better.

src/toxcore/dht/server.rs
kurnevsky: Why not just unit?

kpp: Because later there could be a smart Future Loop that will shutter the time inside `retain` into many pieces.

kurnevsky: Do you really believe this will happen? :)

src/toxcore/dht/server.rs
```diff
@@ -261,114 +255,141 @@ impl Server {
...
+        let payload = packet.get_payload(&self.sk);
+        if payload.is_err() {
             error!("get_payload() fail upon NodesResponse");
-            Box::new( future::err(
+            return Box::new( future::err(
                 Error::new(ErrorKind::Other,
                     "get_payload() fail upon NodesResponse"
             )))
         }
+        let payload = payload.unwrap();
```
kurnevsky: You can use `match` to avoid unwrap here, something like:
```rust
let payload = payload match {
    Ok(payload) => payload,
    Err(_) => { ... }
}
```

src/toxcore/dht/server.rs
```diff
@@ -379,8 +400,18 @@ impl Server {
...
+    fn handle_nat_ping_resp(&self, packet: DhtRequest, payload: NatPingResponse) -> IoFuture<()> {
+        let state = self.state.read();
+        let client = self.get_client(&packet.spk, &state.peers_cache);
+        if client.is_none() {
+            return Box::new( future::err(
+                Error::new(ErrorKind::Other,
+                    "get_client() fail upon NatPingResponse"
+            )))
+        }
+        let client = client.unwrap();
```
kurnevsky: You can use `match` to avoid unwrap here, see above.

src/toxcore/dht/server.rs
kpp: Yes

src/toxcore/dht/server.rs
kpp: Actually he took it from my code. Well... @NamsooCho you may use this trick, but `Err` first. And it is up to you.

kurnevsky: Using unwrap is a bad style when it's avoidable :)

src/toxcore/dht/server.rs
```diff
@@ -170,70 +186,54 @@ impl Server {
...
-                    .map_err(|e| {
-                        e
-                    });
+                let payload = packet.get_payload(&self.sk);
+                let payload = match payload {
+                    Err(_) => {
+                        return Box::new( future::err(
+                            Error::new(ErrorKind::Other,
+                                format!("DhtRequest::get_payload failed with: {:?}",
+                                    payload.unwrap_err())
```
kurnevsky: No need to use unwrap_err here - you already match payload, so you can use Err(err).

kpp:
Rebase/squash the code.

kpp:
Not only Travis CI, but even on my laptop `cargo test` fails.

kpp:
Even on my laptop `cargo test` hangs forever

NamsooCho:
In nightly it compile.
In stable it hangs.

kpp:
I would say in hangs both on stable and nightly but is succeeds on beta.

src/toxcore/dht/server.rs
```diff
@@ -282,13 +297,13 @@ impl Server {
...
     {
-        let state = self.state.read();
+        let mut state = self.state.write();
         let payload = packet.get_payload(&self.sk);
         let payload = match payload {
-            Err(e) => {
+            Err(_) => {
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
-                        format!("PingResponse::get_payload() failed: {:?}", e)
```
kpp: ???

src/toxcore/dht/server.rs
```diff
@@ -301,17 +316,19 @@ impl Server {
             )))
         }
         let client = self.get_client(&packet.pk, &state.peers_cache);
-        let client = match client {
+        let mut client = match client {
             None => {
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
-                        "get_client() failed in handle_ping_resp()"
```
kpp: ???

src/toxcore/dht/server.rs
```diff
@@ -329,10 +346,11 @@ impl Server {
 
         let payload = packet.get_payload(&self.sk);
         let payload = match payload {
-            Err(e) => {
+            Err(_) => {
+                error!("get_payload() fail upon NodesRequest");
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
-                        format!("NodesRequest::get_payload() failed: {:?}", e)
```
kpp: ???

src/toxcore/dht/server.rs
```diff
@@ -360,10 +378,11 @@ impl Server {
         let mut state = self.state.write();
         let payload = packet.get_payload(&self.sk);
         let payload = match payload {
-            Err(e) => {
+            Err(_) => {
+                error!("get_payload() fail upon NodesResponse");
                 return Box::new( future::err(
                     Error::new(ErrorKind::Other,
-                        format!("NodesResponse::get_payload() failed: {:?}", e)
```
kpp: ???

src/toxcore/dht/server.rs
```diff
@@ -742,6 +761,10 @@ mod tests {
         let mut state = alice.state.write();
         state.kbucket = Kbucket::new(KBUCKET_BUCKETS, &alice.pk);
     }
+    // fn add_to_kbucket(alice: &Server, pn: PackedNode) -> bool {
```
kpp: ???

PR #75, opened 2018-03-31
Rename IsStored to AnnounceStatus

kurnevsky:


src/toxcore/onion/packet/announce_response.rs
```diff
@@ -149,7 +149,7 @@ Length   | Content
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct AnnounceResponsePayload {
     /// Variable that represents result of sent `AnnounceRequest`
-    pub is_stored: IsStored,
+    pub is_stored: AnnounceStatus,
```
kpp: should the field be named as `is_stored`?

kurnevsky: I'd left is_stored for field name because c-toxcore and docs use this name - to have persistent terminology. For enum it's less important because they don't have such enum, they just use numbers.
You think we should rename it as well?

kpp: `is_` prefix is good only for binary data that can be either true or false. So yes, rename it, but comment it for the future generations that `your_field` is `is_stored` in c code.

src/toxcore/onion/packet/mod.rs
```diff
@@ -278,11 +278,11 @@ mod tests {
         }
     );
 
-    encode_decode_test!(is_stored_failed, IsStored::Failed);
+    encode_decode_test!(is_stored_failed, AnnounceStatus::Failed);
```
kpp: Don't forget to rename tests

PR #81, opened 2018-04-03
chore(cargo): update minival version of get_if_addrs

kpp:


Cargo.toml
```diff
@@ -39,7 +39,7 @@ tokio = "0.1"
 tokio-io = "0.1"
 nom = "3.2"
 cookie-factory = "0.2.2"
-get_if_addrs = "0.5"
+get_if_addrs = "0.5.1"
```
kurnevsky: I you want to specify minimum version it should be "^0.5.1".

kpp: Why caret? I may use `"0.5"`, call `cargo update` and it will get the latest update (0.5.1). So I don't think we need it.

kurnevsky: ok

PR #94, opened 2018-04-10
Remove Kbucket size limit

kurnevsky:


src/toxcore/dht/kbucket.rs
```diff
@@ -367,16 +352,10 @@ impl Kbucket {
     /** Return the possible internal index of [`Bucket`](./struct.Bucket.html)
         where the key could be inserted/removed.
 
-    Returns `Some(index)` if [`kbucket index`](./fn.kbucket_index.html) is
```
kpp: Why did you remove this doc?

kurnevsky: Because now it should be like "Returns `Some(index)` if [`kbucket index`](./fn.kbucket_index.html) is defined" which means the same as next comment line.

kpp: It is not the same comment. You may return Some(index) when it is 12pm or when you may see the Moon.

kurnevsky: Whereas kbucket_index can't return Some(index) when it sees the Moon for sure :)
It's just the same comment that kbucket_index function has. I can add "only if" words for both of them if you want (but it seems to me that it's quite obvious).

kpp: Let the docs be verbose.

PR #96, opened 2018-04-11
DHT client refactoring

kurnevsky:
This removes Client struct from dht module.


src/toxcore/dht/client.rs
```diff
@@ -24,380 +24,41 @@ Hold infomation of a peer.
...
-        });
-        let lan_sender = ip_addrs.iter().map(|&addr|
-            self.send_to(SocketAddr::new(addr, 33445), lan_packet.clone()) // 33445 is default port for tox
-        );
-
-        let lan_stream = stream::futures_unordered(lan_sender).then(|_| Ok(()));
-        Box::new(lan_stream.for_each(|()| Ok(())))
+    /// set new random ping id to the client and return it
+    pub fn new_ping_id(&mut self) -> u64 {
+        self.ping_id = random_u64();
```
kpp: random + 1, because ping_id must not be equal to 0

kpp:
Still we got bug `ERROR 2018-04-11T09:05:41Z: dht_node: failed to handle packet: Custom { kind: Other, error: StringError("NodesResponse.ping_id does not match") }`. Anyway, good job!

kurnevsky:
Yes, I didn't fix anything regarding ping_id, just refactored. It might happen when we send ping request while already waiting nodes response, but need to investigate anywhay.

src/toxcore/dht/server.rs
```diff
@@ -819,6 +819,15 @@ mod tests {
     const ONION_RETURN_2_PAYLOAD_SIZE: usize = ONION_RETURN_2_SIZE - NONCEBYTES;
     const ONION_RETURN_3_PAYLOAD_SIZE: usize = ONION_RETURN_3_SIZE - NONCEBYTES;
 
+    macro_rules! unmatch {
```
kpp: unpack?

src/toxcore/dht/server.rs
```diff
@@ -414,21 +483,23 @@ impl Server {
     */
     fn handle_nat_ping_req(&self, packet: DhtRequest, payload: NatPingRequest, addr: SocketAddr) -> IoFuture<()> {
         let state = self.state.read();
-        let client = self.create_client(&addr, packet.spk, &state.peers_cache);
 
         if packet.rpk == self.pk { // the target peer is me
-            let resp_payload = NatPingResponse {
+            let resp_payload = DhtRequestPayload::NatPingResponse(NatPingResponse {
```
kpp: Do we have to send nat ping response to ourselves? May be we should not even send nat ping requests to us?

src/toxcore/dht/server.rs
```diff
@@ -947,225 +901,344 @@ mod tests {
...
 
     // handle nat ping response
     #[test]
     fn server_handle_nat_ping_resp_test() {
         let (alice, precomp, bob_pk, _bob_sk, _rx, addr) = create_node();
+
+        // success case
+        let nat_res = NatPingResponse { id: 42 };
+        let nat_payload = DhtRequestPayload::NatPingResponse(nat_res);
+        let dht_req = DhtPacket::DhtRequest(DhtRequest::new(&precomp, &alice.pk, &bob_pk, nat_payload.clone()));
```
kpp: No need to clone nat_payload

src/toxcore/dht/server.rs
kurnevsky: Here we check if someone else sent request to us. If so we handle this request, otherwise - resend to target peer.
It's not correct to check it here though, I've fixed it in other branch.

PR #99, opened 2018-04-12
feat(dht): add more bootstrap nodes to dht example

kpp:


examples/dht_node.rs
```diff
@@ -58,38 +58,61 @@ fn main() {
...
+        // fluke571
+        ("3CEE1F054081E7A011234883BC4FC39F661A55B73637A5AC293DDF1251D9432B", "194.249.212.109:33445"),
+        // MAH69K
+        ("DA4E4ED4B697F2E9B000EEFE3A34B554ACD3F45F5C96EAEA2516DD7FF9AF7B43", "185.25.116.107:33445"),
+        // clearmartin
+        ("CD133B521159541FB1D326DE9850F5E56A6C724B5B8E5EB5CD8D950408E95707", "46.101.197.175:443"),
+        // tastytea
+        ("2B2137E094F743AC8BD44652C55F41DFACC502F125E99E4FE24D40537489E32F", "5.189.176.217:5190"),
+
+    ].iter()
```
kurnevsky: Perhaps here will be clippy warning? To use & instead of iter().

kpp: I checked. There is no any clippy warning.

PR #107, opened 2018-04-19
Ensure that generated onion packets have correct length

kurnevsky:


kpp:
`pub const MAX_DHT_PACKET_SIZE: usize = 512;`.

kurnevsky:
And it's not correct, we will have to increase this limit.
Anyway all onion packets have separate limit of 1400 bytes so this pr makes sence.

kpp:
This PR does make little sense since DhtCodec uses buffer with 512 bytes. So there no way to pass even 1399 bytes with onion packet

kurnevsky:
And as I already said it's not correct. We will have to increase this limit. And I believe it must be much more than 1400 bytes.

kpp:
Then increase it in the current PR.

kurnevsky:
Increased to 2048, c-toxcore has the same limit:
```c
#define MAX_UDP_PACKET_SIZE 2048
```

PR #113, opened 2018-04-25
Add sender's address to onion sink

kurnevsky:


kpp:
update tcp_server example

kurnevsky:
What do you mean by update? Currently tcp_server example doesn't create onion sink so it doesn't have to be updated.

kpp:
Alright

PR #117, opened 2018-04-27
Nodes request example

NamsooCho:


examples/dht_node.rs
```diff
@@ -214,16 +230,16 @@ fn add_nat_sender(base_selector: IoFuture<()>, server_obj: &Server, pk: PublicKe
...
         }))
 }
-fn add_nodes_sender(base_selector: IoFuture<()>, server_obj: &Server) -> IoFuture<()> {
+fn add_nodes_sender(base_selector: IoFuture<()>, server_obj: &mut Server) -> IoFuture<()> {
     // 20 seconds for NodesRequest
-    let interval = Duration::from_secs(20);
+    let interval = Duration::from_secs(1);
     let nodes_wakeups = Interval::new(Instant::now() + interval, interval);
-    let server_obj_c = server_obj.clone();
+    let mut server_obj_c = server_obj.clone();
```
kurnevsky: Server can't be mutable because it's cloned, original server won't be changed.

examples/dht_node.rs
```diff
@@ -216,14 +232,14 @@ fn add_nat_sender(base_selector: IoFuture<()>, server_obj: &Server, pk: PublicKe
 }
 fn add_nodes_sender(base_selector: IoFuture<()>, server_obj: &Server) -> IoFuture<()> {
```
kurnevsky: You should rename this function accordingly.

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,377 @@
...
+
+use toxcore::dht::packed_node::*;
+use toxcore::dht::kbucket::*;
+use toxcore::crypto_core::*;
+use toxcore::dht::server::*;
+use toxcore::dht::server::client::*;
+use toxcore::io_tokio::IoFuture;
+
+/// Maximum number of close nodes per friend.
+pub const MAX_CLOSEST_NODES_PER_FRIEND: u8 = 8;
```
kurnevsky: It seems it's not used?

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,377 @@
...
+            bootstrap_times: Arc::new(RwLock::new(bootstrap_times)),
+            bootstrap_nodes: Arc::new(RwLock::new(Bucket::new(None))),
+        }
+    }
+
+    /// send NodesRequest packet to bootstap_nodes, close list
+    pub fn send_nodes_req_packets(&self, server: &Server,
+                                  ping_interval: u64, nodes_req_interval: u64, bad_node_timeout: u64) -> IoFuture<()> {
+        let ping_bootstrap_nodes = self.ping_bootstrap_nodes(server);
+        let ping_and_get_close_nodes = {
```
kurnevsky: No need in brackets.

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,372 @@
...
+    last_nodes_req_time: Arc<RwLock<Instant>>,
+    /// Counter for bootstappings.
+    bootstrap_times: Arc<RwLock<u32>>,
+    /// Nodes to bootstrap.
+    bootstrap_nodes: Arc<RwLock<Bucket>>,
+}
+
+impl DhtFriend {
+    /// Create new DhtFriend object
+    pub fn new(pk: PublicKey, bootstrap_times: u32) -> Self {
```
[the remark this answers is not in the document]
NamsooCho: Maximum times is defined 5, and if user want only 2 times, he can set it to 3.

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,372 @@
...
+use toxcore::dht::server::*;
+use toxcore::dht::server::client::*;
+use toxcore::io_tokio::IoFuture;
+
+/// Hold friend related info.
+pub struct DhtFriend {
+    /// Public key of friend
+    pk: PublicKey,
+    /// close nodes of friend
+    close_nodes: Arc<RwLock<Bucket>>,
```
kurnevsky: Do you really need RwLock for all these fields? It seems to me that you can take `&mut self` in all methods.

src/toxcore/dht/dht_friend.rs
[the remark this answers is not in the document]
kurnevsky: ok, then it's not obvious how to set this number, you should at least comment this.

src/toxcore/dht/dht_friend.rs
NamsooCho: In the future, DhtFriend will be used in various case.
Especially close_nodes should be maintained one object even if it is cloned.
I don't know the trade-off of permitting mutable self vs. RwLock.
But &mut self may cause problem in the future.

src/toxcore/dht/dht_friend.rs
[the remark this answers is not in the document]
NamsooCho: Ok

src/toxcore/dht/dht_friend.rs
NamsooCho: Ah. that's right.

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +105,242 @@ pub struct Server {
...
+        Box::new(future::ok(()))
+    }
+
+    /// main loop of dht server, call this function every second
+    pub fn dht_main_loop(&self, kill_timeout: u64, ping_timeout: u64, ping_interval: u64, bad_node_timeout: u64, nodes_req_interval: u64) -> IoFuture<()> {
+        self.remove_timedout_clients(Duration::from_secs(kill_timeout));
+        self.remove_timedout_ping_ids(Duration::from_secs(ping_timeout));
+
+        let mut friends = self.friends.write();
+        let friends = friends.deref_mut();
```
kurnevsky: It seems there is no need in this deref_mut.

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,370 @@
...
+                if client.last_resp_time.elapsed() < Duration::from_secs(bad_node_timeout) {
+                    good_nodes.push(node.clone());
+                }
+            }).collect::<Vec<_>>();
+
+        let res = if !good_nodes.is_empty()
+            && self.last_nodes_req_time.elapsed() >= Duration::from_secs(nodes_req_interval)
+            && self.bootstrap_times < MAX_BOOTSTRAP_TIMES {
+
+            let random_node = good_nodes[random_u32() as usize % good_nodes.len()];
```
kurnevsky: c-toxcore has hack here to send requests more often to close nodes:
```c
if ((num_nodes - 1) != rand_node) {
    rand_node += rand() % (num_nodes - (rand_node + 1));
}
```
Maybe we should do the same?

NamsooCho: done

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,377 @@
...
+                }
+            }).collect::<Vec<_>>();
+
+        let res = if !good_nodes.is_empty()
+            && self.last_nodes_req_time.elapsed() >= Duration::from_secs(nodes_req_interval)
+            && self.bootstrap_times < MAX_BOOTSTRAP_TIMES {
+
+            let num_nodes = good_nodes.len();
+            let mut random_node = random_u32() as usize % num_nodes;
+            if 0 != random_node {
```
kurnevsky: Add comment please that this is needed for increase probability of sending request to a close node, because this is not trivial.

NamsooCho: done

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +105,250 @@ pub struct Server {
...
+
+            nodes_stream.for_each(|()| Ok(()))
+        };
+
+        let res = ping_bootstrap_nodes.join4(ping_and_get_close_nodes,
+                                             send_nodes_req_random,
+                                             send_nodes_req_packets)
+            .map(|((), (), (), ())| ());
+
+        self.bootstrap_nodes.write().deref_mut().nodes.clear();
```
kurnevsky: If you move this to `ping_bootstrap_nodes` function you won't need to obtain lock twice.

examples/dht_node.rs
```diff
@@ -230,16 +237,16 @@ fn add_nat_sender(base_selector: IoFuture<()>, server_obj: &Server, pk: PublicKe
...
-    let interval = Duration::from_secs(20);
+    let interval = Duration::from_secs(1);
     let nodes_wakeups = Interval::new(Instant::now() + interval, interval);
     let server_obj_c = server_obj.clone();
     let nodes_sender = nodes_wakeups
         .map_err(|e| Error::new(ErrorKind::Other, format!("Nodes timer error: {:?}", e)))
         .for_each(move |_instant| {
             println!("nodes_wakeup");
-            server_obj_c.periodical_nodes_req()
+            server_obj_c.dht_main_loop(KILL_NODE_TIMEOUT, PING_TIMEOUT, PING_INTERVAL, BAD_NODE_TIMEOUT, NODES_REQ_INTERVAL)
```
kpp: Better create DhtMainLoopArgs with named fields and pass only 1 arg struct instead of list of args

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,381 @@
...
+        let res = ping_bootstrap_nodes.join3(
+            ping_and_get_close_nodes, send_nodes_req_random
+            ).map(|((),(),())| () );
+
+        Box::new(res)
+    }
+
+    fn ping_bootstrap_nodes(&mut self, server: &Server) -> IoFuture<()> {
+        let mut bootstrap_nodes = Bucket::new(None);
+        mem::swap(&mut bootstrap_nodes, &mut self.bootstrap_nodes);
```
kpp: Better use Option<Bucket> for DhtFriend::bootstrap_nodes

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,381 @@
...
+                server.send_nodes_req(*node, self.pk, client)
+            });
+
+        let nodes_stream = stream::futures_unordered(nodes_sender).then(|_| Ok(()));
+
+        Box::new(nodes_stream.for_each(|()| Ok(())))
+    }
+
+    // ping to close nodes of friend
+    fn ping_and_get_close_nodes(&mut self, server: &Server, ping_interval: u64) -> IoFuture<()> {
```
kpp: Better pass Duration as arg for ping_interval

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,381 @@
...
+        Box::new(nodes_stream.for_each(|()| Ok(())))
+    }
+
+    fn send_nodes_req_random(&mut self, server: &Server, bad_node_timeout: u64, nodes_req_interval: u64) -> IoFuture<()> {
+        let mut good_nodes = Vec::new();
+
+        let mut peers_cache = server.get_peers_cache().write();
+
+        self.close_nodes.nodes.iter()
+            .map(|node| {
```
kpp: use `.filter`

src/toxcore/dht/dht_friend.rs
```diff
@@ -0,0 +1,381 @@
...
+        assert!(friend.bootstrap_nodes.try_add(&friend_pk, &PackedNode {
+            pk: node_pk1,
+            saddr: "127.0.0.1:33445".parse().unwrap(),
+        }));
+        assert!(friend.bootstrap_nodes.try_add(&friend_pk, &PackedNode {
+            pk: node_pk2,
+            saddr: "127.0.0.1:33446".parse().unwrap(),
+        }));
+
+        assert!(friend.send_nodes_req_packets(&server, 0, 0, 0).wait().is_ok());
```
kpp: What is 0, 0, 0?

src/toxcore/dht/dht_friend.rs
kurnevsky: You sure that option is better? In this case we will have to check if it's Some or if it's None when we add new node to the list. If Some - add new node, if None - make it Some and again add new node. Seems to me like one additional if for no reason - empty bucket just means that there is no nodes.
In fact this `mem::swap` was added to mitigate borrow chacker and not to obtain lock twice (in the same code in the Server struct).

src/toxcore/dht/dht_friend.rs
NamsooCho: All done.

src/toxcore/dht/dht_friend.rs
kpp: OK

examples/dht_node.rs
```diff
@@ -110,9 +111,16 @@ fn main() {
...
-        assert!(server_obj.try_add_to_kbucket(&bootstrap_pn));
+        assert!(server_obj.try_add_to_close_nodes(&bootstrap_pn));
     }
 
+    // add example friend
+    let pk_str = "9DA18776D7A8ABED7DB67D9B41B853D099A3D4E73C5925B74759E2CFF6289643";
+    let friend_pk_bytes: [u8; 32] = FromHex::from_hex(&pk_str).unwrap();
+    // create PK from bytes
+    let friend_pk = PublicKey::from_slice(&friend_pk_bytes).unwrap();
+    server_obj.add_friend(DhtFriend::new(friend_pk, 0));
```
kpp: What is `0`?

examples/dht_node.rs
```diff
@@ -230,16 +237,24 @@ fn add_nat_sender(base_selector: IoFuture<()>, server_obj: &Server, pk: PublicKe
...
+    let interval = Duration::from_secs(1);
     let nodes_wakeups = Interval::new(Instant::now() + interval, interval);
     let server_obj_c = server_obj.clone();
     let nodes_sender = nodes_wakeups
         .map_err(|e| Error::new(ErrorKind::Other, format!("Nodes timer error: {:?}", e)))
         .for_each(move |_instant| {
             println!("nodes_wakeup");
-            server_obj_c.periodical_nodes_req()
+            let args = DhtMainLoopArgs {
+                kill_node_timeout: KILL_NODE_TIMEOUT,
```
kpp: You don't have to define these constants in server/mod.rs since in future these args may be set via config file. Just hardcode plain values here

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
+    close_nodes: Arc<RwLock<Kbucket>>,
     // symmetric key used for onion return encryption
     onion_symmetric_key: Arc<RwLock<PrecomputedKey>>,
     // onion announce struct to handle onion packets
     onion_announce: Arc<RwLock<OnionAnnounce>>,
+    // friends vector of dht node
+    friends: Arc<RwLock<Vec<DhtFriend>>>,
+    // nodes vector for bootstrap
+    bootstrap_nodes: Arc<RwLock<Bucket>>,
+    bootstrap_times: Arc<RwLock<u32>>,
```
kpp: Add doc for bootstrap_times

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
-// hold client object connected and kbucket object, this struct object is shared by threads
-struct ServerState {
-    /// store client object which has sent request packet to peer
-    pub peers_cache: HashMap<PublicKey, ClientData>,
-    /// Close List (contains nodes close to own DHT PK)
-    pub kbucket: Kbucket,
+/// Struct for grouping parameters to Server's main loop
+pub struct DhtMainLoopArgs {
+    /// timeout for remove clients in peers_cache
+    pub kill_node_timeout: u64,
```
kpp: Better let them be `Instant`s. Or you may add comment that these values are in seconds

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
         }
     }
+
+    /// return peers_cache member variable
+    pub fn get_peers_cache(&self) -> &Arc<RwLock<HashMap<PublicKey, ClientData>>> {
+        &self.peers_cache
+    }
+
+    /// add friend
+    pub fn add_friend(&self, friend: DhtFriend) -> IoFuture<()> {
```
kpp: Why do you return future from this function?

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
+
+        let mut friends = self.friends.write();
+
+        let ping_bootstrap_nodes = self.ping_bootstrap_nodes();
+        let ping_and_get_close_nodes = self.ping_and_get_close_nodes(Duration::from_secs(args.ping_interval));
+        let send_nodes_req_random = self.send_nodes_req_random(Duration::from_secs(args.bad_node_timeout), Duration::from_secs(args.nodes_req_interval));
+        let send_nodes_req_packets = {
+            let nodes_sender = friends.iter_mut()
+                .map(|friend| {
+                    friend.send_nodes_req_packets(self, Duration::from_secs(args.ping_interval), Duration::from_secs(args.nodes_req_interval), Duration::from_secs(args.bad_node_timeout))
```
kpp: This line is too long

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
+
+            let nodes_stream = stream::futures_unordered(nodes_sender).then(|_| Ok(()));
+
+            nodes_stream.for_each(|()| Ok(()))
+        };
+
+        let res = ping_bootstrap_nodes.join4(ping_and_get_close_nodes,
+                                             send_nodes_req_random,
+                                             send_nodes_req_packets)
+            .map(|((), (), (), ())| ());
```
kpp: `map(|_| ())` would also be fine

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
+
+        Box::new(res)
+    }
+
+    fn ping_bootstrap_nodes(&self) -> IoFuture<()> {
```
kpp: Add doc about what this function is doing

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
+        let mut bootstrap_nodes = Bucket::new(None);
+        mem::swap(&mut bootstrap_nodes, self.bootstrap_nodes.write().deref_mut());
```
kpp: add comments why do you do that

src/toxcore/dht/server/mod.rs
```diff
@@ -93,136 +106,264 @@ pub struct Server {
...
+
+        Box::new(nodes_stream.for_each(|()| Ok(())))
+    }
+
+    fn ping_and_get_close_nodes(&self, ping_interval: Duration) -> IoFuture<()> {
+        let close_nodes = self.close_nodes.read();
+        let mut peers_cache = self.peers_cache.write();
+
+        let nodes_sender = close_nodes.iter()
+            .map(|node| {
```
kpp: use `.filter` by `client.last_ping_req_time.elapsed() >= ping_interval` and then `.map` values and send nodes request for each item

src/toxcore/dht/server/mod.rs
```diff
@@ -491,14 +629,21 @@ impl Server {
...
                 // not worried about removing evicted nodes from peers_cache
                 // they will be removed by timeout eventually since we won't
                 // ping them anymore
-                state.kbucket.try_add(node);
+                close_nodes.try_add(node);
+                bootstrap_nodes.try_add(&self.pk, node);
+                friends.iter_mut()
+                    .map(|friend| {
+                        friend.add_to_close(node);
+                    }).collect::<Vec<_>>();
```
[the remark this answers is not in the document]

kurnevsky:
@NamsooCho could you please rebase and fix clippy warnings?

NamsooCho:
Ok. wait some minutes

NamsooCho:
rebasing and removing clippy warnings is done.

PR #121, opened 2018-04-28
refactor(dht): use constant keypair in example dht node

kpp:


kurnevsky:
There will be conflict if several people run this example simultaneously with the same keys.

NamsooCho:
But this is needed to test NatPingRequest/NatPingResponse.

kpp:
@NamsooCho Well... right. I will push the branch so you may test it locally. Also we may add a special flag like `--const_keypair` to switch from random key to the specialized one.

PR #128, opened 2018-05-03
refactor(dht): use constant keypair in example dht node

kpp:
usage: `$ cargo run --example dht_node -- --const_key`


kurnevsky:
It's ok, but in the future we should save generated keys pair in a file and then load it.

PR #131, opened 2018-05-08
Net crypto packets

kurnevsky:


src/toxcore/dht/packet/cookie.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+}
+
+impl EncryptedCookie {
+    /// Create `EncryptedCookie` from `Cookie` encrypting in with `symmetric_key`
+    pub fn new(symmetric_key: &secretbox::Key, payload: Cookie) -> EncryptedCookie {
+        let nonce = &secretbox::gen_nonce();
+        let mut buf = [0; 88];
+        let (_, size) = payload.to_bytes((&mut buf, 0)).unwrap();
+        let payload = secretbox::seal(&buf[..size], nonce, symmetric_key);
```
kpp: Better create `let nonce = secretbox::gen_nonce();` and pass `&nonce` here

src/toxcore/dht/packet/cookie.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// DHT `PublicKey`
+    pub dht_pk: PublicKey,
+}
+
+impl FromBytes for Cookie {
+    named!(from_bytes<Cookie>, do_parse!(
+        time: be_u64 >>
+        real_pk: call!(PublicKey::from_bytes) >>
+        dht_pk: call!(PublicKey::from_bytes) >>
+        eof!() >>
```
kpp: Why eof?

src/toxcore/dht/packet/crypto_handshake.rs
```diff
@@ -0,0 +1,203 @@
...
+    pub payload: Vec<u8>
+}
+
+impl FromBytes for CryptoHandshake {
+    named!(from_bytes<CryptoHandshake>, do_parse!(
+        tag!("\x1a") >>
+        cookie: call!(EncryptedCookie::from_bytes) >>
+        nonce: call!(Nonce::from_bytes) >>
+        payload: take!(248) >>
+        eof!() >>
```
kpp: why eof?

kurnevsky: c-toxcore checks exact length - if it's bigger then packet will be droppet. So I think we can be restrictive here too - it's better to drop suspicious packet then process them.

src/toxcore/dht/packet/cookie.rs
kurnevsky: c-toxcore checks exact length - if it's bigger then packet will be droppet. So I think we can be restrictive here too - it's better to drop suspicious packet then process them.

kpp: ok

src/toxcore/dht/packet/crypto_data.rs
```diff
@@ -0,0 +1,243 @@
...
+`1`      | `0x1b`
+`2`      | Last 2 bytes of the `Nonce`
+variable | Payload
+
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct CryptoData {
+    /// Last two bytes of `Nonce` for the current encrypted payload in BigEndian
+    /// format
+    pub nonce_last_bytes: u16,
```
kpp: I don't like that we lose 2 bytes of Nonce. Shall we alarm Toktok?

PR #138, opened 2018-05-17
Add BootstrapInfo handler

NamsooCho:


kurnevsky:
As I remember only bootstrap nodes handle this packet. Does c-toxcore use this to check if it's bootstrap node or usual client?

src/toxcore/dht/server/mod.rs
```diff
@@ -1114,6 +1118,14 @@ impl Server {
         });
         self.send_to(packet.ip_port.to_saddr(), next_packet)
     }
+    /// handle BootstrapInfo, respond with BootstrapInfo with version = 07032018
+    fn handle_bootstrap_info(&self, _packet: BootstrapInfo, addr: SocketAddr) -> IoFuture<()> {
+        let packet = DhtPacket::BootstrapInfo(BootstrapInfo {
+            version: 07032018,
```
kurnevsky: I think it's better to move version to constant, so we can easily find and update it.

NamsooCho: Ah! I forgot it

NamsooCho:
For now, we do not distinguish bootstrap node or usual client.
Maybe this handler will be moved to bootstrap node only module.

kpp:
> /** Sent by both client and server, only server will respond.
When server receives this packet it may respond with the version of the library
plus MoTD (message of the day). The max length of MoTD is 256 bytes so the max packet
lenght of server BootstrapInfo is 261=(1+4+256) bytes.

> Client must send a BootstrapInfo of exactly 78 bytes, the only 1 field is required: `packet type`
which is filled automatically. So version may be filled with any value, so does MoTD, but
it has to be exactly 73=(78-1-4) bytes long. The server should check that the size of the
packet is exactly 78 bytes long (or MoTD=73 bytes filled with any values). Frankly speaking,
there should be neither `version` nor `motd` fields in the request version, the serialized form
should be 1 byte with packet type + (78-1) bytes of trash, but this implementation is simplified.

src/toxcore/dht/server/mod.rs
```diff
@@ -1114,6 +1119,14 @@ impl Server {
         });
         self.send_to(packet.ip_port.to_saddr(), next_packet)
     }
+    /// handle BootstrapInfo, respond with BootstrapInfo with version = 07032018
```
kurnevsky: Better not to mention exact version here - we can forget to update it.

src/toxcore/dht/server/mod.rs
```diff
@@ -2636,4 +2661,12 @@ mod tests {
...
         assert!(alice.dht_main_loop(args).wait().is_ok());
     }
+
+    #[test]
+    fn server_set_bootstrap_info_test() {
+        let (mut alice, _precomp, _bob_pk, _bob_sk, _rx, _addr) = create_node();
+
+        alice.set_bootstrap_info(42, "test".as_bytes().to_owned());
+        assert_eq!(alice.tox_core_version, 42);
+    }
```
kpp: missing test for MODT

PR #142, opened 2018-05-21
Net crypto handshake

kurnevsky:


examples/dht_node.rs
```diff
@@ -79,7 +81,15 @@ fn main() {
     // Create a channel for server to communicate with network
     let (tx, rx) = mpsc::unbounded::<(DhtPacket, SocketAddr)>();
 
+    let tx_c = tx.clone();
+    let dht_send_fn = Arc::new(move |packet: DhtPacket, addr: SocketAddr|
```
kpp: You could have give a copy of `tx` to NetCrypto instead of `dht_send_fn`. Why so complicated?

src/toxcore/net_crypto/net_crypto.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /// The main loop that should be run at least 20 times per second
+    pub fn main_loop(&self) -> IoFuture<()> {
+        let connections = self.connections.read();
+        let mut send_futures = Vec::new();
+        // Only one cycle over all connections to prevent many lock acquirements
+        for connection in connections.values() {
+            let mut connection = connection.write();
+            let send_future = self.send_status_packet(connection.deref_mut());
```
kpp: Do you really have to call deref_mut here?

src/toxcore/net_crypto/net_crypto.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub fn remove_timed_out(&self) {
+        self.connections.write().retain(|_, connection| !connection.read().is_timed_out());
+    }
+
+    /// The main loop that should be run at least 20 times per second
+    pub fn main_loop(&self) -> IoFuture<()> {
+        let connections = self.connections.read();
+        let mut send_futures = Vec::new();
+        // Only one cycle over all connections to prevent many lock acquirements
+        for connection in connections.values() {
```
kpp: connections.values().iter map collect?

examples/dht_node.rs
kurnevsky: c-toxcore has a lot of callbacks in `net_crypto` module, I'm not sure how we will deal with them. It was just an attempt to use callbacks instead of stream for each callback. But in this particular case you are right - it's easier to pass `tx` directly.

src/toxcore/net_crypto/net_crypto.rs
kurnevsky: No so easy. If change this code to `map` with closure this closure will require `&self` argument to live longer then method body:
```
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter in function call due to conflicting requirements
```

examples/dht_node.rs
```diff
@@ -79,7 +80,10 @@ fn main() {
     // Create a channel for server to communicate with network
     let (tx, rx) = mpsc::unbounded::<(DhtPacket, SocketAddr)>();
 
+    let net_crypto = NetCrypto::new(tx.clone(), server_pk, server_sk.clone());
+
     let mut server_obj = Server::new(tx, server_pk, server_sk);
+    server_obj.set_net_crypto(net_crypto);
```
kpp: When won't the Server have NetCrypto? is it possible?

src/toxcore/net_crypto/net_crypto.rs
```diff
@@ -0,0 +1,219 @@
...
+        for (&pk, connection) in connections.iter() {
+            let mut connection = connection.write();
+
+            if connection.is_timed_out() {
+                timed_out.push((pk, connection.udp_addr));
+                continue;
+            }
+
+            let send_future = self.send_status_packet(connection.deref_mut());
+            send_futures.push(send_future);
```
kpp: use connections.iter().map collect instead of mutable send_futures

kurnevsky: The only way to rewrite this cycle to map:
```rust
        let mut timed_out = Vec::new();
        // Only one cycle over all connections to prevent many lock acquirements
        let send_futures = connections.iter().map(|(&pk, connection)| {
            (pk, connection.write())
        }).filter(|&(pk, ref connection)| {
            if connection.is_timed_out() {
                timed_out.push((pk, connection.udp_addr));
                false
            } else {
                true
            }
        }).map(|(_, mut connection)| {
            self.send_status_packet(connection.deref_mut())
        }).collect::<Vec<_>>();
```
It has two `map`s and dirty `filter` with side effect. You sure that it worth it?

examples/dht_node.rs
kurnevsky: It might be possible when we start only bootstrap node, without the messenger. But actually it can be created with `NetCrypto` as well.

kpp: Got it

examples/dht_node.rs
```diff
@@ -76,10 +77,23 @@ fn main() {
         gen_keypair()
     };
 
+    let (real_pk, _real_sk) = gen_keypair();
+
     // Create a channel for server to communicate with network
     let (tx, rx) = mpsc::unbounded::<(DhtPacket, SocketAddr)>();
 
+    // Create a channel for DHT PublicKey updates
```
kpp: Please add more comments on what is going on

src/toxcore/dht/server/mod.rs
```diff
@@ -129,6 +130,9 @@ pub struct Server {
     // should be redirected to TCP sender trough this sink
     // None if there is no TCP relay
     tcp_onion_sink: Option<TcpOnionTx>,
+    // Net crypto module that handles `CookieRequest`, `CookieResponse`,
+    // `CryptoHandshake` and `CryptoData` packets
+    net_crypto: Option<NetCrypto>,
```
kpp: Please mark that it could be None in case of pure bootstrap server

kurnevsky:
Yes.

src/toxcore/net_crypto/crypto_connection.rs
```diff
@@ -0,0 +1,393 @@
...
+/// direct UDP connection is considered dead
+pub const UDP_DIRECT_TIMEOUT: u64 = 8;
+
+/// Packet that should be sent to the peer every second together with info how
+/// many times it was sent and when it was sent last time
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StatusPacket {
+    /// Packet that should be sent every second. Depending on `ConnectionStatus`
+    /// it can be `CookieRequest` or `CryptoHandshake`
+    pub packet: DhtPacket,
```
kpp: We'd better have a enum of `CookieRequest` with `CryptoHandshake` instead of DhtPacket

src/toxcore/net_crypto/crypto_connection.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub fn is_udp_alive(&self) -> bool {
+        self.udp_received_time
+            .and_then(|time| time.elapsed().ok())
+            .map(|duration| duration < Duration::from_secs(UDP_DIRECT_TIMEOUT))
+            .unwrap_or(false)
+    }
+
+    /// Check if we should send UDP packet regardless of whether UDP is dead or
+    /// alive. In this case we shouldn't rely on UDP only and send the same
+    /// packet via TCP relay
```
kpp: then it should be renamed to `tcp_attempt_should_be_made`?

kpp:
Everything is ok except for 2 marks

src/toxcore/net_crypto/crypto_connection.rs
kurnevsky: No, there is 2 possible cases - UDP is dead or UDP is alive. If it's dead we will always send packet via TCP. But in some cases we can send packet via UDP as well. So the name is correct - it checks if we should send UDP packet together with TCP packet.

PR #144, opened 2018-05-23
add Good and Bad status of node

NamsooCho:


src/toxcore/dht/server/mod.rs
```diff
@@ -168,12 +170,38 @@ impl Server {
...
             motd: Vec::new(),
+            config: ConfigArgs::default(),
             tcp_onion_sink: None
         }
     }
 
+    /**
+    Create new `Server` instance with TCP onion.
+    */
+    pub fn new_with_tcp_onion(tx: Tx, pk: PublicKey, sk: SecretKey, tcp_onion_sink: TcpOnionTx) -> Server {
```
kurnevsky: This function is removed from master - it's replaced with `set_tcp_onion_sink`.

src/toxcore/dht/dht_node.rs
```diff
@@ -18,35 +18,70 @@
...
-impl ClientData {
-    /// create ClientData object
-    pub fn new() -> ClientData {
-        ClientData {
+impl DhtNode {
+    /// create DhtNode object
+    pub fn new() -> DhtNode {
+        DhtNode {
+            pk : gen_keypair().0,
+            saddr: "127.0.0.1:33445".parse().unwrap(),
```
kurnevsky: It's better to pass `pk` and `saddr` as parameters because in current state these field can be used only in tests.

src/toxcore/dht/dht_friend.rs
```diff
@@ -89,9 +89,10 @@ impl DhtFriend {
 
         let mut peers_cache = server.get_peers_cache().write();
 
-        let nodes_sender = bootstrap_nodes.nodes.iter()
+        let bootstrap_nodes = bootstrap_nodes.to_packed_node();
+        let nodes_sender = bootstrap_nodes.iter()
             .map(|node| {
-                let client = peers_cache.entry(node.pk).or_insert_with(ClientData::new);
+                let client = peers_cache.entry(node.pk).or_insert_with(|| DhtNode::new(*node));
```
kurnevsky: Use just `or_insert` method in such cases, no need to wrap it into closure.

NamsooCho:
```
let client = peers_cache.entry(node.pk).or_insert_with(DhtNode::new(*node));
    the trait `std::ops::FnOnce<()>` is not implemented for `toxcore::dht::dht_node::DhtNode

Comment (kurnevsky) [reply src/toxcore/dht/dht_friend.rs]: Not `or_insert_with` but [`or_insert`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert).

Comment (NamsooCho) [reply src/toxcore/dht/dht_friend.rs]: done

File: src/toxcore/dht/kbucket.rs [hunk 4]
```diff
@@ -412,11 +537,13 @@ impl Kbucket {
         let mut bucket = Bucket::new(Some(4));
         for buc in &*self.buckets {
             for node in &*buc.nodes {
-                bucket.try_add(pk, node);
+                bucket.try_add(pk, &node.clone().into());
             }
         }
         trace!("Returning nodes: {:?}", &bucket.nodes);
-        bucket.nodes
+        bucket.nodes.iter()
```
kurnevsky: Use `to_packed_node` method.

src/toxcore/dht/kbucket.rs
```diff
@@ -259,6 +350,54 @@ impl Bucket {
...
+    Returns `true` if [`Bucket`] where node could be placed is not full
+    and node is not already in the [`Bucket`].
+
+    Otherwise `false` is returned.
+
+    [`Bucket`]: ./struct.Bucket.html
+    [`PackedNode`]: ./struct.PackedNode.html
+    */
+    pub fn can_add(&mut self, base_pk: &PublicKey, new_node: &PackedNode) -> bool {
+        let mut new_node: DhtNode = new_node.clone().into();
```
kurnevsky: Why clone?

src/toxcore/dht/dht_node.rs
```diff
@@ -18,35 +18,71 @@
...
+Generate ping_id on request packet, check ping_id on response packet.
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DhtNode {
+    /// Socket addr of node.
+    pub saddr: SocketAddr,
+    /// Public Key of the node.
+    pub pk: PublicKey,
+    /// Status of node
+    pub status: NodeStatus,
```
kurnevsky: It seems we can have method that returns status instead of storing it as a field because this status uniquely defined by `last_resp_time`.

src/toxcore/dht/kbucket.rs
```diff
@@ -175,11 +255,22 @@ impl Bucket {
         trace!(target: "Bucket", "With bucket: {:?}; PK: {:?} and new node: {:?}",
             self, base_pk, new_node);
 
-        match self.nodes.binary_search_by(|n| base_pk.distance(&n.pk, &new_node.pk)) {
+        let mut new_node: DhtNode = new_node.clone().into();
```
kurnevsky: Why clone?

src/toxcore/dht/kbucket.rs
NamsooCho: Ok. PackedNode implements Copy trait

src/toxcore/dht/kbucket.rs
```diff
@@ -259,6 +293,45 @@ impl Bucket {
...
+    */
+    pub fn can_add(&mut self, base_pk: &PublicKey, new_node: &PackedNode) -> bool {
+        let new_node: DhtNode = (*new_node).into();
+
+        match self.nodes.binary_search_by(|n| base_pk.replace_order(n, &new_node, self.bad_node_timeout)) {
+            Ok(_index) => false, // node already exist in bucket, so can't add node
+            Err(index) if index == self.nodes.len() => { // can't find node in bucket
+                if self.is_full() { // bucket is full, so can't add node
+                    false
+                } else { // buceket has space, so can add node
```
kpp: buceket

src/toxcore/dht/kbucket.rs
```diff
@@ -259,6 +293,45 @@ impl Bucket {
...
+                    true
+                }
+            },
+            Err(_index) => true, // node is not found in bucket, so can add node
+        }
+    }
+
+    /// convert voctor of DhtNode to vector of PackedNode
```
kpp: voctor

NamsooCho: done

src/toxcore/dht/kbucket.rs
```diff
@@ -412,11 +471,12 @@ impl Kbucket {
         let mut bucket = Bucket::new(Some(4));
         for buc in &*self.buckets {
             for node in &*buc.nodes {
-                bucket.try_add(pk, node);
+                bucket.try_add(pk, &node.clone().into());
```
kurnevsky: No need to clone.

src/toxcore/dht/kbucket.rs
```diff
@@ -259,6 +293,45 @@ impl Bucket {
...
+
+    Returns `true` if [`Bucket`] where node could be placed is not full
+    and node is not already in the [`Bucket`].
+
+    Otherwise `false` is returned.
+
+    [`Bucket`]: ./struct.Bucket.html
+    [`PackedNode`]: ./struct.PackedNode.html
+    */
+    pub fn can_add(&mut self, base_pk: &PublicKey, new_node: &PackedNode) -> bool {
```
kurnevsky: Why `&mut`?

src/toxcore/dht/kbucket.rs
```diff
@@ -442,11 +502,11 @@ impl Kbucket {
     [`Bucket`]: ./struct.Bucket.html
     [`PackedNode`]: ./struct.PackedNode.html
     */
-    pub fn can_add(&self, pk: &PublicKey) -> bool {
-        match self.bucket_index(pk) {
+    pub fn can_add(&mut self, new_node: &mut PackedNode) -> bool {
```
kurnevsky: Why `&mut`?

src/toxcore/dht/kbucket.rs
```diff
@@ -485,7 +550,7 @@ impl Iterator for KbucketIter {
             match self.buckets[self.pos_b].nodes.get(self.pos_pn) {
                 Some(s) => {
                     self.pos_pn += 1;
-                    return Some(*s);
+                    return Some(s.clone().into());
```
kurnevsky: No need to clone.

kurnevsky:
One more thing - I don't like that now we store thse same `DhtNode` in two different places - `kbucket` and `nodes_cache`. We need `nodes_cache` only for pings (as ping_array in c-toxcore). But now we store one same node twice - inside `nodes_cache` and inside `kbucket`. But for me it's ok if it's done in another pr.

src/toxcore/dht/kbucket.rs
NamsooCho: done

NamsooCho:
I agree.
DhtNode should be used in kbucket.
ClientData which is removed for now should be used in peers_cache to check ping_id.

src/toxcore/dht/kbucket.rs
```diff
@@ -970,17 +1029,14 @@ mod tests {
...
-                if kbucket.try_add(node) {
-                    let index = kbucket_index(&pk, &node.pk);
-                    // none of nodes with the same index can be added
-                    // to the kbucket
-                    assert!(pns.iter()
-                        .filter(|pn| kbucket_index(&pk, &pn.pk) == index)
-                        .all(|pn| !kbucket.can_add(&pn.pk)));
+            kbucket.set_bad_node_timeout(10);
+
+            for mut node in pns.clone() {
```
kurnevsky: No need in clone?

NamsooCho: done

examples/dht_server.rs
```diff
@@ -228,30 +228,35 @@ fn add_server_main_loop(base_selector: IoFuture<()>, server_obj: &Server) -> IoF
...
-                    nodes_req_interval: 20,
-                    nat_ping_req_interval: 3,
+                    nodes_req_interval: 0,
+                    nat_ping_req_interval: 0,
                 };
 
-                server_obj_c.dht_main_loop(args)
-            } else {
-                bootstrap_fast = true;
+                server_obj_c.set_config_values(args);
```
kpp: No need to set config values each wakeup. You may do that before you start the server

kurnevsky: Btw, do we need to have these as variables? Perhaps we can just use constants? This will require to change `Instant` to `SystemTyme` for the tests, but there are some benefits of this change anyway. So is there some scenario where setting these custom (not default) variables might be usefull?

NamsooCho: Yes. bootstrap_fast is a flag for first calling of dht_main_loop().
The separation of first call and other calls is for fast bootstrapping.
So set config is called just two times( first call, and second call)

NamsooCho: I also think about that. These constants may not changed per copy of toxcore.

src/toxcore/dht/kbucket.rs
```diff
@@ -131,7 +158,8 @@ impl Bucket {
                 trace!("Creating a new Bucket with default capacity.");
                 Bucket {
                     capacity: BUCKET_DEFAULT_SIZE as u8,
-                    nodes: Vec::with_capacity(BUCKET_DEFAULT_SIZE)
+                    nodes: Vec::with_capacity(BUCKET_DEFAULT_SIZE),
+                    bad_node_timeout: Duration::default(),
```
kpp: Should we set a preconfigured constant here?

kpp:
good job!

src/toxcore/dht/kbucket.rs
NamsooCho: bad_node_timeout settting is done using this function 
https://github.com/tox-rs/tox/blob/45c6414fdd11c3c4fe7ada99aea981679885eaec/src/toxcore/dht/kbucket.rs#L326
And Bucket::new() is used many place in toxcore.
To have consistency, we have to use current new(Option<u8>) prototype.

kurnevsky: @kpp  was speaking about `bad_node_timeout`. Now we have ConfigArgs and set this value from these args. But actually it should be constant, and for now you can use this constant as default value instead of `Duration::default()`.

kpp: > . But actually it should be constant, and for now you can use this constant as default value instead of Duration::default().

true

examples/dht_server.rs
kpp: resolved

kpp:
>  In the pr https://github.com/tox-rs/tox/pull/144 we have same types for kbucket and for peers_cache. Thus we store a lot of redundant information.

@kurnevsky what do you mean?

kurnevsky:
@kpp I mean that now we store info about nodes in two places: (kbucket)[https://github.com/tox-rs/tox/blob/master/src/toxcore/dht/server/mod.rs#L110] and (peers_cache)[https://github.com/tox-rs/tox/blob/master/src/toxcore/dht/server/mod.rs#L108]. In kbucket we store address and pk of node, in peers_cache - ping_id, last_resp_time, last_ping_req_time. In this pr both kbucket and peers_cache store the same struct - `DhtNode`. So for one node we will store this info twice.
But we can't just remove peers_cache because it plays the role of ping_array - firstly node gets to peers_cache and only after receiving ping response it gets to kbucket.

src/toxcore/dht/kbucket.rs
NamsooCho: done

kpp:
I don't think it is a problem right now.

PR #146, opened 2018-05-31
chore(readme): update the status of the project

kpp:


kurnevsky:
Actually `CookieResponse` and `CryptoHandshake` are not handled properly yet. To do it we will need module for friends handling to initiate and accept incoming connections.

kurnevsky:
No, it's not like client/server. It's used to establish connection between two equal peers. We just need to know with whom we should establish this connection. And from whom accept connection. To handle this we need to have list of friends with already known dht key (for that we need onion client) and initiate connection (create CryptoConnection with known dht and long term keys).

kpp:
ok

kurnevsky:
But CookieRequest is fully handled now, so status can be updated :)

kpp:
))))

PR #147, opened 2018-06-01
separate ClientData from DhtNode

NamsooCho:
DhtNode has redundancy in peers_cache.
So, separate ClientData from DhtNode.


kurnevsky:
Maybe we can choose a better name for `ClientData` and `peers_cache`? Because now it doesn't show the real purpose of this struct. Perhaps something like `PingData` and `ping_map`?

NamsooCho:
done

PR #150, opened 2018-06-03
CryptoData handling

kurnevsky:


src/toxcore/net_crypto/mod.rs
```diff
@@ -85,18 +116,20 @@ pub struct NetCrypto {
...
+    connections: Arc<RwLock<HashMap<PublicKey, Arc<RwLock<CryptoConnection>>>>>,
     /// Long term keys by IP address of DHT node map. `SocketAddr` can't be used
     /// as a key since it contains additional info for `IPv6` address.
     keys_by_addr: Arc<RwLock<HashMap<(IpAddr, /*port*/ u16), PublicKey>>>,
 }
 
 impl NetCrypto {
     /// Create new `NetCrypto` object
-    pub fn new(udp_tx: UdpTx, dht_pk_tx: DhtPkTx, dht_pk: PublicKey, dht_sk: SecretKey, real_pk: PublicKey) -> NetCrypto {
+    pub fn new(udp_tx: UdpTx, dht_pk_tx: DhtPkTx, lossless_tx: LosslessTx, lossy_tx: LossyTx, dht_pk: PublicKey, dht_sk: SecretKey, real_pk: PublicKey) -> NetCrypto {
```
NamsooCho: Too many arguments. Is there any other way?

src/toxcore/net_crypto/mod.rs
```diff
@@ -300,6 +337,151 @@ impl NetCrypto {
...
+        let mut n = 1;
+
+        for i in send_array.buffer_start .. send_array.buffer_end {
+            if data.len() == 0 {
+                break
+            }
+
+            if n == data[0] {
+                if let Some(packet) = send_array.get_mut(i) {
+                    if packet.sent_time.elapsed() > rtt {
```
NamsooCho: Why do you use Duration instead of SystemTime.

src/toxcore/net_crypto/mod.rs
kurnevsky: Right, and there will be even more args. I don't know what to do in this case. Perhaps we can group some args in structs.

src/toxcore/net_crypto/packets_array.rs
```diff
@@ -0,0 +1,224 @@
...
+fn real_index(index: u32) -> usize {
+    (index % CRYPTO_PACKET_BUFFER_SIZE) as usize
+}
+
+/// Deque-like struct for packets queue that allows random writings by the
+/// packet index
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct PacketsArray<T> {
+    /// Packets buffer of `CRYPTO_PACKET_BUFFER_SIZE` length
+    pub buffer: Vec<Option<Box<T>>>,
```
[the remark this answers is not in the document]
kurnevsky: Stack overflow has nothing to do with Box here. It's needed to reduce memory usage, since Option<Box<T>> is just a pointer this buffer will use maximum 8 * CRYPTO_PACKET_BUFFER_SIZE = 8 * 32768 = 256kb memory. Without box it will use much more.
The reason why buffer is preallocated is that we can receive packets in random order so we need random access to the buffer to write them. It's possible to implement the same structure using something like `VecDequeue` but it will be more difficult (since we will need to push many empty elements to write at random index). So It's better to stick to a simpler implementation now and then change it when it will be covered with tests (if we will decide to change it at all).

src/toxcore/net_crypto/mod.rs
kurnevsky: Only one thing I can suggest at the moment - create struct `NetCryptoNewArgs` and pass it to the `new` function.

kpp: > create struct NetCryptoNewArgs and pass it to the new function.

Good

examples/dht_server.rs
```diff
@@ -88,12 +88,33 @@ fn main() {
     // DHT PublicKey as well.
     let (dht_pk_tx, dht_pk_rx) = mpsc::unbounded();
 
+    let (lossless_tx, lossless_rx) = mpsc::unbounded();
```
kpp: Describe what it is.

examples/dht_server.rs
```diff
@@ -200,6 +221,8 @@ fn main() {
     let server = add_server_main_loop(server, &server_obj);
     let server = add_onion_key_refresher(server, &server_obj);
     let server = server.join(dht_pk_handler).map(|_| ());
+    let server = server.join(lossless_handler).map(|_| ());
```
kpp: Why did you chose join instead of select?

src/toxcore/dht/packet/crypto_data.rs
```diff
@@ -163,6 +163,17 @@ impl ToBytes for CryptoDataPayload {
     }
 }
 
+impl CryptoDataPayload {
+    /// Get data skipping zeros at the front
```
kpp: Why do we need to skip zeros?

src/toxcore/dht/packet/crypto_data.rs
```diff
@@ -163,6 +163,17 @@ impl ToBytes for CryptoDataPayload {
...
+    pub fn real_data(&self) -> Vec<u8> {
+        self.data
+            .iter()
+            .cloned()
```
kpp: Do we really need to clone it once more?

kurnevsky: Because `iter()` returns iterator over `&u8` and I need just value `u8`. And `cloned()` method just clones inner value from reference.

src/toxcore/dht/packet/crypto_data.rs
kurnevsky: Because data can be padded with any number of zeros at the beginning. Only after zeros will be packet id and actual payload.

examples/dht_server.rs
kurnevsky: https://docs.rs/futures/0.2.1/futures/trait.FutureExt.html#method.join - Joins the result of two futures, waiting for them both to complete.
https://docs.rs/futures/0.2.1/futures/trait.FutureExt.html#method.select - Waits for either one of two differently-typed futures to complete.

I need both futures but not any.

kurnevsky: And I think that `select` in other places should be changed to `join` so we will wait all the futures. Actually they are equal here since all these futures will never be complete but logically `join` fits better here.

src/toxcore/dht/packet/crypto_data.rs
kpp: Can we do that in packet parsing?

kurnevsky: Right, just checked - the length of this padding is defined by the length of the data. So we can remove it when parsing and add it back when writing to a byte array.

PR #151, opened 2018-06-04
add Ping iterator

NamsooCho:
Current implementation of Ping is not correct.
This PR is fixing it.

### DIFF
--- a/src/toxcore/dht/kbucket.rs
+++ b/src/toxcore/dht/kbucket.rs
@@ -401,6 +401,15 @@ impl Kbucket {
         )
     }
 
+    /// return peer which has pk
+    pub fn find_node(&self, pk: &PublicKey) -> Option<&DhtNode> {
+        self.bucket_index(pk).and_then(|index|
+            self.buckets[index]
+                .find(&self.pk, pk)
+                .map(|node_index| &self.buckets[index].nodes[node_index])
+        )
+    }
+
     /** Return the possible internal index of [`Bucket`](./struct.Bucket.html)
         where the key could be inserted/removed.
 
--- a/src/toxcore/dht/server/mod.rs
+++ b/src/toxcore/dht/server/mod.rs
@@ -27,7 +27,7 @@ This module works on top of other modules.
 */
 
 pub mod client;
-pub mod ping;
+pub mod ping_sender;
 pub mod hole_punching;
 
 use futures::{Future, Sink, Stream, future, stream};
@@ -56,7 +56,7 @@ use toxcore::io_tokio::*;
 use toxcore::dht::dht_friend::*;
 use toxcore::dht::server::hole_punching::*;
 use toxcore::tcp::packet::OnionRequest;
-use toxcore::dht::server::ping::*;
+use toxcore::dht::server::ping_sender::*;
 use toxcore::net_crypto::NetCrypto;
 
 /// Shorthand for the transmit half of the message channel.
--- a/src/toxcore/dht/server/mod.rs
+++ b/src/toxcore/dht/server/mod.rs
@@ -26,8 +26,9 @@ Functionality needed to work as a DHT node.
 This module works on top of other modules.
 */
 
-pub mod hole_punching;
 pub mod client;
+pub mod ping;
+pub mod hole_punching;
 
 use futures::{Future, Sink, Stream, future, stream};
 use futures::sync::mpsc;
@@ -55,6 +56,7 @@ use toxcore::io_tokio::*;
 use toxcore::dht::dht_friend::*;
 use toxcore::dht::server::hole_punching::*;
 use toxcore::tcp::packet::OnionRequest;
+use toxcore::dht::server::ping::*;
 use toxcore::net_crypto::NetCrypto;
 
 /// Shorthand for the transmit half of the message channel.
@@ -122,6 +124,7 @@ pub struct Server {
     // setting this value to 0 will do sending 5 times
     bootstrap_times: Arc<RwLock<u32>>,
     last_nodes_req_time: Arc<RwLock<Instant>>,
+    ping_sender: Arc<RwLock<Ping>>,
     // toxcore version used in BootstrapInfo
     tox_core_version: u32,
     // message used in BootstrapInfo
@@ -152,6 +155,8 @@ pub struct ConfigArgs {
     pub bad_node_timeout: u64,
     /// interval in seconds for random NodesRequest
     pub nodes_req_interval: u64,
+    /// interval in seconds for iteration of sending PingRequest
+    pub ping_iter_interval: u64,
     /// interval in seconds for NatPingRequest
     pub nat_ping_req_interval: u64,
 }
@@ -175,6 +180,7 @@ impl Server {
             bootstrap_nodes: Arc::new(RwLock::new(Bucket::new(None))),
             bootstrap_times: Arc::new(RwLock::new(0)),
             last_nodes_req_time: Arc::new(RwLock::new(Instant::now())),
+            ping_sender: Arc::new(RwLock::new(Ping::new())),
             tox_core_version: 0,
             motd: Vec::new(),
             config: ConfigArgs::default(),
@@ -219,30 +225,43 @@ impl Server {
         let ping_and_get_close_nodes = self.ping_and_get_close_nodes(Duration::from_secs(self.config.ping_interval));
         let send_nodes_req_random = self.send_nodes_req_random(Duration::from_secs(self.config.bad_node_timeout),
                                                                Duration::from_secs(self.config.nodes_req_interval));
-        let send_nodes_req_packets = {
-            let mut friends = self.friends.write();
+        let send_nodes_req_to_friends = self.send_nodes_req_to_friends();
 
-            let nodes_sender = friends.iter_mut()
-                .map(|friend| {
-                    friend.send_nodes_req_packets(self, Duration::from_secs(self.config.ping_interval),
-                                                  Duration::from_secs(self.config.nodes_req_interval),
-                                                  Duration::from_secs(self.config.bad_node_timeout))
-                });
-
-            let nodes_stream = stream::futures_unordered(nodes_sender).then(|_| Ok(()));
-
-            nodes_stream.for_each(|()| Ok(()))
-        };
+        let ping_sender = self.iterate_ping_list(Duration::from_secs(self.config.ping_iter_interval));
 
         let send_nat_ping_req = self.send_nat_ping_req(Duration::from_secs(self.config.nat_ping_req_interval));
 
-        let res = ping_bootstrap_nodes.join5(ping_and_get_close_nodes,
-                                             send_nodes_req_random,
-                                             send_nodes_req_packets,
-                                             send_nat_ping_req)
+        let res = future::join_all(vec![ping_bootstrap_nodes,
+                                        ping_and_get_close_nodes,
+                                        send_nodes_req_random,
+                                        send_nodes_req_to_friends,
+                                        ping_sender,
+                                        send_nat_ping_req])
             .map(|_| ());
 
         Box::new(res)
+    }
+
+    // send PingRequest using Ping object
+    fn iterate_ping_list(&self, ping_send_interval: Duration) -> IoFuture<()> {
+        let mut ping_sender = self.ping_sender.write();
+
+        ping_sender.iterate_ping_list(&self, ping_send_interval)
+    }
+
+    // send NodesRequest to friends
+    fn send_nodes_req_to_friends(&self) -> IoFuture<()> {
+        let mut friends = self.friends.write();
+
+        let nodes_sender = friends.iter_mut()
+            .map(|friend| {
+                friend.send_nodes_req_packets(self, Duration::from_secs(self.config.ping_interval),
+                                              Duration::from_secs(self.config.nodes_req_interval),
+                                              Duration::from_secs(self.config.bad_node_timeout))
+            });
+
+        let nodes_stream = stream::futures_unordered(nodes_sender).then(|_| Ok(()));
+        Box::new(nodes_stream.for_each(|()| Ok(())))
     }
 
     // send NodesRequest to nodes gotten by NodesResponse
@@ -352,18 +371,6 @@ impl Server {
         Box::new( future::ok(()) )
     }
 
-    /// send PingRequest to all peers in close_nodes
-    pub fn send_pings(&self) -> IoFuture<()> {
-        let close_nodes = self.close_nodes.read();
-        let ping_sender = close_nodes.iter().map(|peer| {
-            self.send_ping_req(&peer)
-        });
-
-        let pings_stream = stream::futures_unordered(ping_sender).then(|_| Ok(()));
-
-        Box::new(pings_stream.for_each(|()| Ok(())))
-    }
-
     /// Send PingRequest to node
     pub fn send_ping_req(&self, node: &PackedNode) -> IoFuture<()> {
         let mut ping_map = self.ping_map.write();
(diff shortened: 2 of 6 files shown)


src/toxcore/dht/kbucket.rs
```diff
@@ -401,6 +401,15 @@ impl Kbucket {
         )
     }
 
+    /// return peer which has pk
+    pub fn find_node(&self, pk: &PublicKey) -> Option<DhtNode> {
+        self.bucket_index(pk).and_then(|index|
+            self.buckets[index]
+                .find(&self.pk, pk)
+                .map(|node_index| self.buckets[index].nodes[node_index].clone())
```
kurnevsky: I's better to return reference here. Then you can clone after calling this function if you really need it.

NamsooCho: done

src/toxcore/dht/server/mod.rs
```diff
@@ -219,32 +225,45 @@ impl Server {
...
+                                        send_nodes_req_to_friends,
+                                        ping_sender,
+                                        send_nat_ping_req])
             .map(|_| ());
 
         Box::new(res)
     }
 
+    // send PingRequest using Ping object
+    fn iterate_ping_list(&self, ping_send_interval: Duration) -> IoFuture<()> {
```
kpp: Function named as `iterate_` must return iterator. Better name it like `send_pings`

src/toxcore/dht/server/mod.rs
```diff
@@ -219,32 +225,45 @@ impl Server {
...
+        let mut ping_sender = self.ping_sender.write();
+
+        ping_sender.iterate_ping_list(&self, ping_send_interval)
+    }
+
+    // send NodesRequest to friends
+    fn send_nodes_req_to_friends(&self) -> IoFuture<()> {
```
kpp: It is OK for now but better create a new PR with only 1 change that includes this function. 1 PR per 1 feature.

src/toxcore/dht/server/mod.rs
```diff
@@ -462,6 +469,14 @@ impl Server {
         match packet {
             DhtPacket::PingRequest(packet) => {
                 debug!("Received ping request");
+                // send PingRequest
+                let node_to_ping = PackedNode {
+                    pk: packet.pk,
+                    saddr: addr,
+                };
+
+                self.ping_sender.write().try_add(&self, &node_to_ping);
```
kpp: `try_add` must return `Result` or boolean. Better name it as `add`.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Tox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+
+/*!
+Module for PingRequest
```
kpp: Add a better doc

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+            return Box::new(future::ok(()))
+        }
+
+        let mut now = Instant::now();
+        let mut nodes_to_send_ping = Bucket::new(None);
+
+        mem::swap(&mut self.last_time_send_ping, &mut now);
+        mem::swap(self.nodes_to_send_ping.write().deref_mut(), &mut nodes_to_send_ping);
+
+        if nodes_to_send_ping.is_empty() {
```
kpp: Useless check

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+use toxcore::dht::dht_node::*;
+use toxcore::io_tokio::IoFuture;
+
+/// Hold data for sending PingRequest
+pub struct Ping {
+    last_time_send_ping: Instant,
+    nodes_to_send_ping: RwLock<Bucket>,
+}
+
+impl Ping {
```
kpp: It is `PingSender` actually

src/toxcore/dht/server/mod.rs
```diff
@@ -462,6 +469,14 @@ impl Server {
         match packet {
             DhtPacket::PingRequest(packet) => {
                 debug!("Received ping request");
+                // send PingRequest
```
kpp: Do it inside handle_ping_req

src/toxcore/dht/server/mod.rs
```diff
@@ -470,6 +485,14 @@ impl Server {
             },
             DhtPacket::NodesRequest(packet) => {
                 debug!("Received NodesRequest");
+                // send PingRequest
```
kpp: Do it inside handle_nodes_req

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+        server.friends.read().iter()
+            .map(|friend| friend.close_nodes.nodes.iter().any(|peer| peer.pk == node.pk))
+            .any(|result_bool| result_bool)
+    }
+
+    fn is_in_ping_list(node: &PackedNode, ping_list: &Bucket) -> bool {
+        ping_list.nodes.iter().any(|peer| peer.pk == node.pk)
+    }
+
+    fn is_iterate_time(&self, iterate_interval: Duration) -> bool {
```
kpp: rename to can_send_pings

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+        server.friends.read().iter().any(|friend| friend.pk == node.pk)
+    }
+
+    fn is_in_close_list(node: &PackedNode, server: &Server) -> bool {
+        server.friends.read().iter()
+            .map(|friend| friend.close_nodes.nodes.iter().any(|peer| peer.pk == node.pk))
+            .any(|result_bool| result_bool)
+    }
+
+    fn is_in_ping_list(node: &PackedNode, ping_list: &Bucket) -> bool {
```
kpp: It would be better if this function accepted `self`.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+        // if node already exists in ping list, then don't add
+        if Ping::is_in_ping_list(node, nodes_to_send_ping.deref()) {
+            return false
+        }
+
+        // PingRequest is sent only for maximum 8 nodes in Bucket
+        nodes_to_send_ping.try_add(&server.pk, node)
+    }
+
+    /// send PingRequest to all nodes in list
```
kpp: Let it be clear: do we send only 8 pings? And we do it only once since you used mem::swap. Am I correct?

src/toxcore/dht/server/ping.rs
kurnevsky: I wouldn't agree - there may be cases when self is mutable borrowed so we can't pass it as an argument and can't use simple functions that don't require it.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+use toxcore::dht::packed_node::*;
+use toxcore::dht::kbucket::*;
+use toxcore::dht::server::*;
+use toxcore::dht::dht_node::*;
+use toxcore::io_tokio::IoFuture;
+
+/// Hold data for sending PingRequest
+pub struct Ping {
+    last_time_send_ping: Instant,
+    nodes_to_send_ping: RwLock<Bucket>,
```
kurnevsky: Do you really need `RwLock` here? You already have one in the Server.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// send PingRequest to all nodes in list
+    pub fn iterate_ping_list(&mut self, server: &Server, iterate_interval: Duration) -> IoFuture<()> {
+        if !self.is_iterate_time(iterate_interval) {
+            return Box::new(future::ok(()))
+        }
+
+        let mut now = Instant::now();
+        let mut nodes_to_send_ping = Bucket::new(None);
+
+        mem::swap(&mut self.last_time_send_ping, &mut now);
```
kurnevsky: Why do you need to swap it? Just set it to `last_time_send_ping`.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+        mem::swap(self.nodes_to_send_ping.write().deref_mut(), &mut nodes_to_send_ping);
```
kurnevsky: Perhaps [mem::replace](https://doc.rust-lang.org/beta/std/mem/fn.replace.html) will be a bit nicer here.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// return true if node is added, false otherwise
+    pub fn try_add(&mut self, server: &Server, node: &PackedNode) -> bool {
+        // if node already exists in close list and not timed out, then don't send PingRequest
+        match server.close_nodes.read().find_node(&node.pk) {
+            Some(ref node_in_close_list) if !Ping::is_bad_node_timed_out(node_in_close_list, server) => return false,
+            _ => {},
+        };
+
+        // if node is not addable to close list, dont't send PingRequest
+        if !server.close_nodes.read().can_add(node) {
```
kurnevsky: Do not acquire the same lock multiple times.

src/toxcore/dht/server/mod.rs
```diff
@@ -219,32 +225,45 @@ impl Server {
...
-        };
+        let ping_sender = self.iterate_ping_list(Duration::from_secs(self.config.ping_iter_interval));
 
         let send_nat_ping_req = self.send_nat_ping_req(Duration::from_secs(self.config.nat_ping_req_interval));
 
-        let res = ping_bootstrap_nodes.join5(ping_and_get_close_nodes,
-                                             send_nodes_req_random,
-                                             send_nodes_req_packets,
-                                             send_nat_ping_req)
+        let res = future::join_all(vec![ping_bootstrap_nodes,
```
kurnevsky: You can use static array here instead of vector to not allocate the memory.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+        node.last_resp_time.elapsed() > Duration::from_secs(server.config.bad_node_timeout)
+    }
+
+    fn is_friend(node: &PackedNode, server: &Server) -> bool {
+        server.friends.read().iter().any(|friend| friend.pk == node.pk)
+    }
+
+    fn is_in_close_list(node: &PackedNode, server: &Server) -> bool {
+        server.friends.read().iter()
+            .map(|friend| friend.close_nodes.nodes.iter().any(|peer| peer.pk == node.pk))
```
kurnevsky: Use `any` instead of map.

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+
+        if nodes_to_send_ping.is_empty() {
+            return Box::new(future::ok(()))
+        }
+
+        let ping_sender = nodes_to_send_ping.nodes.iter().map(|node| {
+            server.send_ping_req(&(node.clone()).into())
+        });
+
+        let pings_stream = stream::futures_unordered(ping_sender).then(|_| Ok(()));
```
kurnevsky: Why `stream::futures_unordered` but not `future::join_all`?

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,236 @@
...
+impl Ping {
+    /// new Ping object
+    pub fn new() -> Self {
+        Ping {
+            last_time_send_ping: Instant::now(),
+            nodes_to_send_ping: RwLock::new(Bucket::new(None)),
+        }
+    }
+
+    fn is_bad_node_timed_out(node: &DhtNode, server: &Server) -> bool {
```
kurnevsky: Maybe move this function to `DhtNode` and just pass `bad_node_timeout` value there?

src/toxcore/dht/server/ping.rs
NamsooCho: Yes.
ping list is Bucket. So only 8 closer nodes will receive PingReqeust.
Adding node to Bucket and sending Ping is separate jobs.
We check ping Bucket to send Ping every 2 second.

src/toxcore/dht/server/ping.rs
NamsooCho: I want to use entries in current Bucket and then clear the Bucket.

src/toxcore/dht/server/mod.rs
NamsooCho: I don't know how to

src/toxcore/dht/server/ping.rs
NamsooCho: I think it is just my coding style. Is better to use join_all?

src/toxcore/dht/server/ping.rs
NamsooCho: done except some comments which I can't resolve.

src/toxcore/dht/server/ping.rs
kurnevsky: `futures_unordered` returns a stream, but you actually don't need it, you need just a future.

NamsooCho: I used this iteration pattern for a long time. 
And there are many occurrences in DHT using this coding pattern.
Now, I got what you say.
Maybe, I may change all of this coding pattern in another PR.

src/toxcore/dht/server/ping.rs
kurnevsky: Right, and you can do it with replace:
```rust
let nodes_to_send_ping = mem::replace(self.nodes_to_send_ping.write().deref_mut(), Bucket::new(None))
```

src/toxcore/dht/server/ping.rs
kpp: See https://docs.rs/futures/0.1.18/src/futures/stream/futures_unordered.rs.html#56 vs http://rust-lang-nursery.github.io/futures-rs/futures/future/fn.join_all.html .

Better use futures_unordered

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,225 @@
+/*
```
kpp: you forgot to rename file to ping_sender.rs

src/toxcore/dht/server/ping.rs
kpp: I disagree. There may be cases but now there is no one. So please change it to member function

src/toxcore/dht/server/ping.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn is_friend(node: &PackedNode, server: &Server) -> bool {
+        server.friends.read().iter().any(|friend| friend.pk == node.pk)
+    }
+
+    fn is_in_close_list(node: &PackedNode, server: &Server) -> bool {
+        server.friends.read().iter()
+            .any(|friend| friend.close_nodes.nodes.iter().any(|peer| peer.pk == node.pk))
+    }
+
+    fn is_in_ping_list(node: &PackedNode, ping_list: &Bucket) -> bool {
```
kpp: https://github.com/tox-rs/tox/pull/151#discussion_r192849219

src/toxcore/dht/server/ping.rs
kurnevsky: Could you explain why at least?

src/toxcore/dht/server/ping.rs
NamsooCho: done

PR #156, opened 2018-06-13
Do not fall on invalid packets

kurnevsky:
There are two ways to fix handling of incomplete or broken packet:
1. Encode deserialization error type to `Decoder::Item` and then filter out these error. In this approach we should change `type Item = DhtPacket;` to `type Item = Result<DhtPacket, Err>;` in `impl Decoder for DhtCodec` and return `Ok(Err(...))` in `decode` method in case of failure.
2. Add possibility to match on error types so that we could filter out only deserialization errors. This is  a bit more straightforward approach but requires to change error types. The easiest way to do it is using `failure` crate. It will bring some mess with errors in our code but if we are going to customize our errors it could be a good starting point.

This pull request implements the second way. Thoughts?

### DIFF
--- a/examples/dht_server.rs
+++ b/examples/dht_server.rs
@@ -23,6 +23,7 @@
 // an example of DHT node with current code
 //
 extern crate tox;
+extern crate failure;
 extern crate futures;
 extern crate tokio;
 extern crate tokio_io;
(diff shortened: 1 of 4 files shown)


examples/dht_server.rs
```diff
@@ -184,9 +192,9 @@ fn main() {
             }
             sink.send((packet, addr))
         })
+        .map_err(|e| Error::new(ErrorKind::Other, e.compat()))
```
kpp: err_msg?

examples/dht_server.rs
```diff
@@ -155,21 +157,27 @@ fn main() {
...
-        // All tasks must have an `Error` type of `()`. This forces error
-        // handling and helps avoid silencing failures.
-        error!("packet receive error = {:?}", err);
-        Error::new(ErrorKind::Other, "udp receive error")
-    });
+    }).map_err(|e| Error::new(ErrorKind::Other, e.compat()));
 
     let network_writer = rx
-        .map_err(|_| Error::new(ErrorKind::Other, "rx error"))
+        .map_err(|_| err_msg("rx error"))
```
kpp: let it be done in another huge PR when we will move to crate failure to use nice errs instead of string errors

PR #158, opened 2018-06-18
save or load dht daemon status

NamsooCho:


src/toxcore/state_format/mod.rs
```diff
@@ -31,4 +31,4 @@ https://zetok.github.io/tox-spec/#state-format
 
 
 // FIXME: use new dht code instead of old
-// pub mod old;
+pub mod new;
```
kpp: Why did you create a new state_format? Is it compatible to old one?

src/toxcore/dht/daemon_state.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl DaemonState {
+    /// save DHT close list
+    pub fn save(server: &Server) -> Vec<u8> {
+        let nodes = server.close_nodes.read().iter()
+            .map(|node| node.into())
+            .collect::<Vec<PackedNode>>();
+
+        let mut buf = [0u8; 1024 * 1024];
+        if let Ok((_, buf_len)) = DhtState(nodes).to_bytes((&mut buf, 0)) {
```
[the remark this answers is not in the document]
kurnevsky: We can calculate the length of `buf` before calling `to_bytes`. We know the count of `PacketNode`s, we know the size of serialized `PacketNode`, we know overhead that `DhtState` adds - it's all that we need.

src/toxcore/dht/daemon_state.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /// load DHT close list and then re-setup close list
+    pub fn load(server: &Server, saved_data: Vec<u8>) -> IoFuture<()> {
+        let nodes = match DhtState::from_bytes(&saved_data) {
+            IResult::Done(_, DhtState(nodes)) => nodes,
+            _e => return Box::new(
+                future::err(
+                    Error::new(ErrorKind::Other, "Can't find client in ping_map")
```
kurnevsky: Correct error message. You may also include `_e` there - it might contain useful info.

src/toxcore/state_format/old.rs
```diff
@@ -1055,38 +870,110 @@ impl SectionData {
...
-            SectionKind::Friends => Friends::arbitrary(g).to_bytes(),
-            SectionKind::Name => Name::arbitrary(g).0,
-            SectionKind::StatusMsg => StatusMsg::arbitrary(g).0,
-            SectionKind::Status => vec![UserStatus::arbitrary(g) as u8],
-            SectionKind::TcpRelays => TcpRelays::arbitrary(g).to_bytes(),
-            SectionKind::PathNodes => PathNodes::arbitrary(g).to_bytes(),
-            SectionKind::EOF => vec![],
-        };
-        SectionData { kind: kind, data: data }
+        let mut range_val: u8 = 7;
```
kurnevsky: The previous implementation was much cleaner. Why did you rewrite it in a such way?

src/toxcore/state_format/old.rs
```diff
@@ -1055,38 +870,110 @@ impl SectionData {
...
+        (SectionData { kind, data: data.to_vec() })
+    ));
+}
+
+#[cfg(test)]
+impl Arbitrary for SectionKind {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let mut range_val: u8 = 7;
+
+        while range_val > 6 && range_val < 0x0a {
```
kurnevsky: It's just a way to generate random `SectionKind`. Generated value doesn't have to correspond to saved byte. You can assume that generated `7` means `SectionKind::TcpRelays` and `8` means `SectionKind::PathNodes` and remove this cycle.

src/toxcore/dht/daemon_state.rs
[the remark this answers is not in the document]
NamsooCho: Maximum DhtState size will be
Kbucket size = Bucket * 256 = (DhtNode * 8 + 9)  * 256 = (61 * 8 + 9) * 256 = 19,968 bytes.

So, I think that I should modify arbitrary function to produce data which has size within 19,968 bytes.

src/toxcore/state_format/mod.rs
NamsooCho: Old format has the bytes which need for C language. e.g.: null padding and delimiters for separation of fields.
New format do not have section concepts. I think that section concept is not needed.
New format is not compatible with old format.

src/toxcore/state_format/old.rs
NamsooCho: Oh! I am sorry. old.rs is not used.
Now new.rs is included.

src/toxcore/state_format/old.rs
NamsooCho: old.rs is deprecated.

src/toxcore/dht/daemon_state.rs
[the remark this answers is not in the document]
NamsooCho: Now I checked arbitrary for DhtState, and the arbitrary already produce test data which has size  within maximum size.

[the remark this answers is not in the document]
NamsooCho: // kurnevsky 
The buf[] has size of static.
Do you mean that it will be better to use dynamic allocation?

src/toxcore/dht/daemon_state.rs
NamsooCho: done

src/toxcore/state_format/mod.rs
kpp: Why did you do that?

NamsooCho: https://github.com/tox-rs/tox/blob/3c75733792deca27ef42a7655268467999b5d444/src/toxcore/state_format/old.rs#L624-L635

The padding for fixed size section is not good choice(I think). It just increase code's complexity.

https://github.com/tox-rs/tox/blob/3c75733792deca27ef42a7655268467999b5d444/src/toxcore/state_format/old.rs#L1004-L1044

SectionData has no meaning. It also increases code's complexity.

kurnevsky: But with this change we won't be able to load profile saved by `c-toxcore`? Or even migrate it? I'd prefer backward compatible option.

src/toxcore/dht/daemon_state.rs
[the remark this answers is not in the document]
kurnevsky: Then we can use exact constant maximum size? With comment why it's max. 1024 * 1024 is way bigger than we need and it's not obvious why the buffer has this size.

src/toxcore/state_format/mod.rs
NamsooCho: In c-toxcore, the engine itself produces bytes stream, not a saved file.
And take bytes stream as a parameter to load.
How to save these bytes stream to file(s) or how to load these bytes stream from file(s) is up to the user of c-toxcore.
So, two Tox clients which use same c-toxcore can not guarantee it's compatibility.

kurnevsky: In the future if we make backward incompatible changes we should consider using something standardized (like [cbor](http://cbor.io/)) and not inventing a new format.

NamsooCho: Good catch!!!.
I didn't think that our tox-rs may be changed in the future.
And it should provide backward compatibility.

kurnevsky: > How to save these bytes stream to file(s) or how to load these bytes stream from file(s) is up to the user of c-toxcore.
So, two Tox clients which use same c-toxcore can not guarantee it's compatibility.

But this byte stream is defined by toxcore? I mean how it's possible that different clients save the same stream to a file in a different ways and get different incompatible formats?

NamsooCho: Ok. I got what you say.
I will modify this PR to use same format, or to provide backward compatibility at least.

src/toxcore/dht/daemon_state.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /// load DHT close list and then re-setup close list
+    pub fn load(server: &Server, saved_data: Vec<u8>) -> IoFuture<()> {
+        let nodes = match DhtState::from_bytes(&saved_data) {
+            IResult::Done(_, DhtState(nodes)) => nodes,
+            e => return Box::new(
+                future::err(
+                    Error::new(ErrorKind::Other, format!("Can't find client in ping_map: {:?}", e))
```
kurnevsky: This error means that you can't parse `DhtState` from bytes but not find something in ping_map. That is what I meant.

NamsooCho: done

src/toxcore/dht/daemon_state.rs
```diff
@@ -0,0 +1,147 @@
...
+
+impl DaemonState {
+    /// save DHT close list
+    pub fn save(server: &Server) -> Vec<u8> {
+        let nodes = server.close_nodes.read().iter() // DhtNode is reformed to PackedNode through iter()
+            .map(|node| node)
+            .collect::<Vec<PackedNode>>();
+
+        let mut buf = [0u8; DHT_STATE_BUFFER_SIZE];
+        if let Ok((_, buf_len)) = DhtState(nodes).to_bytes((&mut buf, 0)) {
```
kurnevsky: I'd prefer `expect` here since it should be always `Ok`.

NamsooCho: done

src/toxcore/dht/daemon_state.rs
```diff
@@ -0,0 +1,145 @@
...
+        // PackedNode size
+        (
+            32 + // PK size
+            19   // SocketAddr maximum size
+        ) * BUCKET_DEFAULT_SIZE // num of DhtNodes per Bucket : 8
+    ) * KBUCKET_MAX_ENTRIES as usize; // 255
+
+impl DaemonState {
+    /// save DHT close list
+    pub fn save(server: &Server) -> Vec<u8> {
```
kpp: rename to `serialize_old`, because you don't save it somewhere, you do actually serialize it. And `old` corresponds to old C format.

You may note it in the doc comment.

src/toxcore/dht/daemon_state.rs
```diff
@@ -0,0 +1,145 @@
...
+            .collect::<Vec<PackedNode>>();
+
+        let mut buf = [0u8; DHT_STATE_BUFFER_SIZE];
+        let (_, buf_len) = DhtState(nodes).to_bytes((&mut buf, 0)).expect("DhtState(nodes).to_bytes has failed");
+
+        buf[..buf_len].to_vec()
+    }
+
+    /// load DHT close list and then re-setup close list
+    pub fn load(server: &Server, saved_data: Vec<u8>) -> IoFuture<()> {
```
kpp: deserialize_old

kpp:
@NamsooCho  Will you create another PR to actually load/save the state periodically?

src/toxcore/dht/daemon_state.rs
NamsooCho: done

src/toxcore/dht/daemon_state.rs
```diff
@@ -20,9 +20,9 @@
 
 
 /*!
-Save or load states of tox daemon.
-When toxcore start, it load states from saved file.
-Toxcore daemon saves its states to file every 10 minutes.
+serialize or deserialize states of tox daemon.
+When toxcore start, it deserialize states from serialized file.
+Toxcore daemon serializes its states to file every 10 minutes.
```
kpp: > file every 10 minutes

right now it is not true

src/toxcore/dht/daemon_state.rs
```diff
@@ -20,9 +20,9 @@
 
 
 /*!
-Save or load states of tox daemon.
-When toxcore start, it load states from saved file.
-Toxcore daemon saves its states to file every 10 minutes.
+serialize or deserialize states of tox daemon.
+When toxcore start, it deserialize states from serialized file.
```
kpp: > When toxcore start, it deserialize states

When toxcore starts, it deserializes states

src/toxcore/dht/daemon_state.rs
```diff
@@ -65,13 +65,13 @@ impl DaemonState {
...
-    pub fn load(server: &Server, saved_data: Vec<u8>) -> IoFuture<()> {
-        let nodes = match DhtState::from_bytes(&saved_data) {
+    /// deserialize DHT close list and then re-setup close list, old means that the format of deserialization is old version
+    pub fn deserialize_old(server: &Server, serialized_data: Vec<u8>) -> IoFuture<()> {
+        let nodes = match DhtState::from_bytes(&serialized_data) {
             IResult::Done(_, DhtState(nodes)) => nodes,
             e => return Box::new(
                 future::err(
-                    Error::new(ErrorKind::Other, format!("Can't load DHT states from saved file(s) {:?}", e))
+                    Error::new(ErrorKind::Other, format!("Can't deserialize DHT states from serialized file(s) {:?}", e))
```
kpp: > from serialized file

from serialized bytes

NamsooCho: done

PR #159, opened 2018-06-20
chore(deps): update deps of the project

kpp:


src/toxcore/dht/codec.rs
```diff
@@ -88,7 +88,10 @@ impl Decoder for DhtCodec {
 
         match DhtPacket::from_bytes(buf) {
             IResult::Incomplete(needed) => Err(DecodeError::IncompletePacket { len, needed }.into()),
-            IResult::Error(error) => Err(DecodeError::DeserializeError { error }.into()),
+            IResult::Error(error) => {
+                let error_kind = error.into_error_kind();
```
kpp: Although we did not change the version of nom (3.2.1), it changed the enum of IResult.

@kurnevsky do you have any reason why that happened?

kurnevsky: You sure that you need this change? It seems `into_error_kind` is [no-op](https://docs.rs/nom/3.2.1/src/nom/util.rs.html#706-708) - it takes `ErrorKind` and returns it.

src/toxcore/crypto_core.rs
```diff
@@ -58,8 +58,8 @@ pub fn crypto_init() -> bool {
     //       `CRYPTO_INIT` *can't* be modified while it may be read by
     //       something else.
     CRYPTO_INIT_ONCE.call_once(|| {
-        let initialized = ::sodiumoxide::init();
-        unsafe { CRYPTO_INIT = initialized; }
+        ::sodiumoxide::init().expect("could not initialize libsodium");
```
kurnevsky: With this change `crypto_init` will always return `true`? Then better make it unit. btw, it seems we don't need this magic with `call_once` - `::sodiumoxide::init()` can be called multiple times and from different threads:
> sodium_init() initializes the library and should be called before any other function provided by Sodium. It is safe to call this function more than once and from different threads -- subsequent calls won't have any effects.

src/toxcore/dht/codec.rs
kpp: It is a no-op for ErrorKind, whereas IResult::Error contains [Err\<T\>](https://github.com/Geal/nom/blob/1a70c8fa3cbb710c069395e1ee10019a4f8d3709/src/internal.rs#L133)

kurnevsky: Which is synonym to `ErrorKind`: https://docs.rs/nom/3.2.1/nom/simple_errors/type.Err.html

src/toxcore/crypto_core.rs
kpp: Well, https://github.com/sodiumoxide/sodiumoxide/issues/223 . Let's fix it the next sodiumoxide release

src/toxcore/dht/codec.rs
kurnevsky: It was changed to enum only in version 4.

kpp: Oh... I got a strange error that `error` is a type of `Err<&[u8]>`, while `ErrorKind` was expected.

I reverted the code and it worked for me. T_T

kpp:
@kurnevsky 

```
   Compiling libsodium-sys v0.1.0
   Compiling sodiumoxide v0.1.0
   Compiling tox v0.0.4 (file:///home/humbug/zetox)
error[E0308]: mismatched types
  --> src/toxcore/crypto_core.rs:62:32
   |
62 |         unsafe { CRYPTO_INIT = initialized; }
   |                                ^^^^^^^^^^^ expected bool, found enum `std::result::Result`
   |
   = note: expected type `bool`
              found type `std::result::Result<(), ()>`

error[E0308]: mismatched types
  --> src/toxcore/dht/codec.rs:91:74
   |
91 |             IResult::Error(error) => Err(DecodeError::DeserializeError { error }.into()),
   |                                                                          ^^^^^ expected enum `nom::ErrorKind`, found enum `nom::Err`
   |
   = note: expected type `nom::ErrorKind`
              found type `nom::Err<&[u8]>`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0308`.
error: Could not compile `tox`.

To learn more, run the command again with --verbose.
humbug@home:~/zetox$ git diff | cat
diff --git a/Cargo.toml b/Cargo.toml
index 9206765..b4f0b15 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ bytes = "0.4"
 byteorder = "1"
 futures = "0.1"
 log = "0.4"
-sodiumoxide = "0.0.16"
+sodiumoxide = "0.1"
 tokio = "0.1.7"
 tokio-codec = "0.1"
 nom = "3.2.1"
```

PR #160, opened 2018-06-22
Update sodiumoxide to v0.2

kpp:


.travis.yml
```diff
@@ -50,7 +49,6 @@ matrix:
 
     - os: linux
       rust: stable
-      sudo: required
```
kpp: `sudo: required` for kcov. whereas for the root conf and for all other runners you should set `sudo: false`.

README.md
```diff
@@ -28,42 +28,42 @@ issue / pull request should be filled on the reference repository.
...
 [CONTRIBUTING.md](/CONTRIBUTING.md).
 
-## Dependencies
-| **Name** | **Version** |
-|----------|-------------|
-| libsodium | >=1.0.13 |
-
 ## Building
-Fairly simple. You'll need [Rust] >= 1.26.0 and [libsodium].
+Fairly simple. First, install [Rust] >= 1.30.1.
```
kurnevsky: Do we also need clang now?

kurnevsky:
`increment_nonce_number` can use native libsodium function now.

.travis.yml
```diff
@@ -41,6 +36,7 @@ matrix:
   include:
     - os: linux
       rust: nightly
+      sudo: false
```
kurnevsky: Why disable it? And does it have to be disabled twice?

README.md
kpp: No.

kurnevsky: ok, but gcc should be required, shouldn't it? https://github.com/alexcrichton/cc-rs#compile-time-requirements

suhr: Rust actually asks you to install Build Tools for Visual Studio on Windows.

kurnevsky: Don't you think it should be mentioned here?

PR #161, opened 2018-06-23
DHT refactoring

kurnevsky:
This PR:
- implements advanced LAN discovery from `friend_connection.c`
- moves onion key refreshing to `dht_main_loop`
- simplifies dht_server example

I decided to leave LAN discovery wake ups separated from `dht_main_loop` since:
- it'd have required adding additional argument to server
- LAN discovery is optional so if it's disabled we don't need LAN wake ups at all (at least in bootstrap server).

### DIFF
--- a/examples/dht_server.rs
+++ b/examples/dht_server.rs
@@ -312,24 +312,3 @@ fn run_lan_discovery_sender(mut lan_discovery_sender: LanDiscoverySender) -> IoF
         });
     Box::new(future)
 }
-
-fn add_onion_key_refresher(base_selector: IoFuture<()>, server_obj: &Server) -> IoFuture<()> {
-    // Refresh onion symmetric key every 2 hours. This enforces onion paths expiration.
-    let interval = Duration::from_secs(7200);
-    let refresh_onion_key_wakeups = Interval::new(Instant::now() + interval, interval);
-    let server_obj_c = server_obj.clone();
-    let onion_key_updater = refresh_onion_key_wakeups
-        .map_err(|e| Error::new(ErrorKind::Other, format!("Refresh onion key timer error: {:?}", e)))
-        .for_each(move |_instant| {
-            println!("refresh_onion_key_wakeup");
-            server_obj_c.refresh_onion_key();
-            future::ok(())
-        });
-
-    Box::new(base_selector.select(Box::new(onion_key_updater))
-        .map(|_| ())
-        .map_err(move |(err, _select_next)| {
-            error!("Processing ended with error: {:?}", err);
-            err
-        }))
-}
--- a/examples/dht_server.rs
+++ b/examples/dht_server.rs
@@ -226,36 +226,32 @@ fn main() {
             err
         });
 
-    let server: IoFuture<()> = Box::new(network);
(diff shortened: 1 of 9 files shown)


examples/dht_server.rs
```diff
@@ -223,36 +226,32 @@ fn main() {
...
 
-    let server: IoFuture<()> = Box::new(network);
-    let server = add_lan_sender(server, &server_obj, local_addr);
-    let server = add_server_main_loop(server, &server_obj);
-    let server = add_onion_key_refresher(server, &server_obj);
-    let server = server.join(dht_pk_handler).map(|_| ());
-    let server = server.join(lossless_handler).map(|_| ());
-    let server = server.join(lossy_handler).map(|_| ());
+    let server: IoFuture<()> = Box::new(network); // TODO: remove these boxes on rustc 1.26
+    let server: IoFuture<()> = Box::new(server.join(run_server(&server_obj)).map(|_| ()));
```
kpp: use select instead of join

kurnevsky: Could you explain why you think that `select` is should be used here? With `select` you must assume that all these futures are infinite whereas `join` explicitly shows that application will be terminated when all these futures are completed.

kpp: Because if any of the futures resolves/fails, we should return immediately. In your case we will wait until all futures are terminated.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use futures::{Future, Stream};
+
+    macro_rules! unpack {
+        ($variable:expr, $variant:path) => (
```
NamsooCho: How about moving this macro definition to binary_io.rs.
Because this macro is used commonly in many files.

examples/dht_server.rs
```diff
@@ -223,36 +226,32 @@ fn main() {
...
 
-    let nodes_sender = nodes_wakeups
+    let future = dht_wakeups
         .map_err(|e| Error::new(ErrorKind::Other, format!("Nodes timer error: {:?}", e)))
         .for_each(move |_instant| {
-            println!("main_loop_wakeup");
+            trace!("DHT server wake up");
             // flag for fast bootstrapping
             if bootstrap_fast {
                 server_obj_c.dht_main_loop()
```
NamsooCho: Now dht_wakeups starts Instant::now() instead of Instant::now() + interval,
bootstrap_fast is not needed.

examples/dht_server.rs
kurnevsky: But in case if one future resolves with error `join` won't wait other futures.
And another argument for `join` - `select` modifies error type to [either of errors](https://docs.rs/futures/0.1.21/futures/future/struct.Select.html), so I can't just change `join` to `select`, I'll have to add `map_err` for every line here.

kpp: A simple rule:

- use `select` if you don't have to wait for other futures.
- use `join` if you need to.

Do we need to wait for other futures?

PR #162, opened 2018-06-25
Relicense to MIT

kpp:
Any objections?

cc @nokaa, @sudden6, @senia-psm, @quininer, @zetok 

Objections:

- [ ] nokaa
- [ ] sudden6
- [ ] senia-psm
- [ ] quininer
- [ ] zetok


sudden6:
Why? and what code did I commit under GPL here?

kpp:
> Why? and what code did I commit under GPL here?

```
commit af0b6934201b28abde4eba6786d957f6cb1582ba
Author: sudden6 <sudden6@users.noreply.github.com>
Date:   Sun Mar 20 23:07:39 2016 +0100

    docs: fix typo

commit b37a34feff6eb3c47227364888c4240c27f45a27
Author: sudden6 <sudden6@users.noreply.github.com>
Date:   Sun Mar 20 23:02:41 2016 +0100

    docs: Improve documentation of Node

commit c7b987ea812538ee25f63f611f6009ca658326e3
Author: sudden6 <sudden6@users.noreply.github.com>
Date:   Sun Mar 20 22:22:06 2016 +0100

    docs(dht): clarify documentation
```

sudden6:
While I generally prefer GPL, in this case I don't want to object for such a tiny contribution.

Why do you change the license though?

kpp:
>  I don't want to object for such a tiny contribution.

Thank you.

> Why do you change the license though?
There are many reasons. 

- GPL is not allowed in App Store
- MIT code cannot include GPL code, while most of Rust crates are MIT

senia-psm:
No objections. Is it alive?

kpp:
> Is it alive?

Sure! Have a look at the [pulse](https://github.com/tox-rs/tox/pulse/monthly) of the project.

kurnevsky:
Do we need copyright in every file with MIT license? Perhaps we can just enumerate people in readme?

kpp:
> Do we need copyright in every file with MIT license?

You are right. We don't have to. I will fix that.

zetok:
Sorry for late post and being unresponsive, but I object.

> - GPL is not allowed in App Store

That is an artificial problem created by the Apple corp with an aim of making user freedom suffer. Conforming to it by changing license not only doesn't help, but instead it makes things worse, since it clearly signals that things like user freedom are not worth much and can be removed whenever $big_software_platform demands it.

If anything, Apple should change ToS of their App Store.

> - MIT code cannot include GPL code, while most of Rust crates are MIT

This also is not a problem with the license, but only with users of the library who would want to restrict user freedom, which is not allowed with GPL. All that project in question would need to do is to also care about user freedom and be licensed under GPL - technically there is no problem with changing license of the project from MIT to GPL.

Just because majority does something it doesn't mean that being conformant and doing the same thing is the right choice.

---

When I've started this project, I didn't make a mindless choice, but chose GPL after spending time deliberating whether "caveats" of GPL outweigh lack of restrictions in other licenses, like MIT. Outcome is clear.

GPL not only provides user freedom, but also includes mechanisms that would preserve this freedom at later time.

With this in mind, I hoped that even if one day I were no longer developing the project, the user freedom would get preserved, be it by people developing the library itself, or by future consumers of the library.


---

Since I've already talked about the present and the past, what about the future? Although I'm not adamant about changing license, there should be good reasons for change and actual benefits from changing license, rather than demerits like removal of user freedom protection mechanisms. And a few other protection mechanisms that MIT doesn't include, including, but not limited to patents.



---

Ah, and I just had a funny thought: "Hmm, would I sell my soul to Satan if he offered to allow me in his app store? Tempting, but I dunno…"  :D

kpp:
> Sorry for late post and being unresponsive, but I object.

Alright. Let's resolve this issue.

kpp:
My SSD crashed, so this is a new ToxID: 153694973E1898DC06944AB3DB36A6161EACD1520B175779C0DE0E10E5940B265D9CFA620270

zetok:
Sent invite, my ID remains the same: 29AE62F95C56063D833024B1CB5C2140DC4AEB94A80FF4596CACC460D7BAA062E0A92C3424A0

zetok:
So, a sentence from conversation that pretty much sums up the conclusion:

> I think that dual-licensing [with MIT] would be acceptable as long as GPL stays, and to ensure that a reason for using GPL needs to be clearly stated in licenses/README.md, namely patent protection

PR #164, opened 2018-06-26
Move run functions to core

kurnevsky:


examples/dht_server.rs
```diff
@@ -241,64 +239,3 @@ fn main() {
...
-
-    let future = dht_wakeups
-        .map_err(|e| Error::new(ErrorKind::Other, format!("Nodes timer error: {:?}", e)))
-        .for_each(move |_instant| {
-            trace!("DHT server wake up");
-            // flag for fast bootstrapping
-            if bootstrap_fast {
-                server_obj_c.dht_main_loop()
-            } else {
-                bootstrap_fast = true;
```
kpp: You missed why Namsoo played with `bootstrap_fast` and config params. Run dht server and you will see that the bootstrap happens only on the ~20 second.

kurnevsky: Yes, but it's not the best way to fix this delay, and I'm not sure that 20s delay worth it. I thought about making `last_nodes_req_time` variable optional with `None` initial value so we can see that we should send request immediately. But left this to another pr, since I think we should review and refactor bootstrap process - it has a lot to improve.

kpp: > I'm not sure that 20s delay worth this juggling with config params

20s delay is too much.

kurnevsky: ok, made it optional. It seems it's not the best fix either, but butter than this magic with params.

examples/dht_server.rs
```diff
@@ -227,8 +225,8 @@ fn main() {
         });
 
     let server: IoFuture<()> = Box::new(network); // TODO: remove these boxes on rustc 1.26
-    let server: IoFuture<()> = Box::new(server.select(run_server(&server_obj)).map(|_| ()).map_err(|(e, _)| e));
-    let server: IoFuture<()> = Box::new(server.select(run_lan_discovery_sender(lan_discovery_sender)).map(|_| ()).map_err(|(e, _)| e));
+    let server: IoFuture<()> = Box::new(server.select(server_obj.run()).map(|_| ()).map_err(|(e, _)| e));
```
kpp: Remove `Box::new` if it is posssible

kurnevsky: As TODO says it's possible only with rustc 1.26. There was a fix for huge compilation time with long list of inner generics. And we have on travis minimal rustc 1.25.

kpp: OK

PR #167, opened 2018-06-29
make LanDiscovery optional

NamsooCho:
The work to separate LanDiscovery is already done by Evegeny.
I make processing of LanDiscovery optional.


src/toxcore/dht/server/mod.rs
```diff
@@ -882,6 +889,11 @@ impl Server {
     */
     fn handle_lan_discovery(&self, packet: LanDiscovery, addr: SocketAddr) -> IoFuture<()> {
         // if Lan Discovery packet has my PK, then it is sent by myself.
```
kpp: You mixed up comments.

PR #172, opened 2018-07-10
Poll future in ping_sender.rs

NamsooCho:


src/toxcore/dht/server/ping_sender.rs
```diff
@@ -64,7 +64,11 @@ impl PingSender {
 
         // If node is friend and don't exist in friend's close list then send PingRequest
         if PingSender::is_friend(node, server) && !PingSender::is_in_close_list(node, server) {
-            server.send_ping_req(node);
+            match server.send_ping_req(node).wait() {
```
kpp: Which will give us a blocking code. Please find another way to fix it.

NamsooCho: done

PR #174, opened 2018-07-10
DHT refactoring

kurnevsky:


src/toxcore/dht/daemon_state.rs
```diff
@@ -34,8 +34,8 @@ pub const DHT_STATE_BUFFER_SIZE: usize =
 impl DaemonState {
     /// serialize DHT states, old means that the format of seriaization is old version
     pub fn serialize_old(server: &Server) -> Vec<u8> {
-        let nodes = server.close_nodes.read().iter() // DhtNode is reformed to PackedNode through iter()
-            .map(|node| node)
+        let nodes = server.close_nodes.read().iter()
+            .map(|node| node.clone().into())
```
kpp: use `iter().cloned()`

src/toxcore/dht/dht_friend.rs
```diff
@@ -104,19 +99,19 @@ impl DhtFriend {
     }
 
     // send NodesRequest to random node which is in close list
-    fn send_nodes_req_random(&mut self, server: &Server, bad_node_timeout: Duration, nodes_req_interval: Duration) -> IoFuture<()> {
-        let mut ping_map = server.get_ping_map().write();
+    fn send_nodes_req_random(&mut self, server: &Server) -> IoFuture<()> {
+        if self.last_nodes_req_time.map_or(false, |time| clock_elapsed(time) < Duration::from_secs(NODES_REQ_INTERVAL)) &&
+            self.bootstrap_times >= MAX_BOOTSTRAP_TIMES {
+            return Box::new(future::ok(()));
```
kpp: add comment smt like this: too many bootstrap nodes for the current round. try again a little bit later

src/toxcore/dht/dht_node.rs
```diff
@@ -14,57 +14,31 @@ use std::ops::Sub;
...
-                    NodeStatus::Bad => { // Bad, Bad
-                        self.distance(&node1.pk, &node2.pk)
-                    },
-                }
-            },
+        if node1.is_bad() {
+            if node2.is_bad() {
+                self.distance(&node1.pk, &node2.pk) // both bad
+            } else {
+                Ordering::Greater // good, bad
```
kpp: you missed the order of 'bad' and 'good' in comments.

src/toxcore/dht/dht_node.rs
```diff
@@ -14,57 +14,31 @@ use std::ops::Sub;
...
+            }
+        } else {
+            if node2.is_bad() {
+                Ordering::Less // bad, good
```
kpp: fix the order in comments

src/toxcore/dht/dht_node.rs
```diff
@@ -114,22 +90,14 @@ impl DhtNode {
...
-            NodeStatus::Bad
-        } else {
-            NodeStatus::Good
+            last_ping_req_time: None,
         }
     }
 
     /// check it the node is timed out
-    pub fn is_bad_node_timed_out(&self, server: &Server) -> bool {
-        self.calc_status(Duration::from_secs(server.config.bad_node_timeout)) == NodeStatus::Bad
```
kpp: Add comment: the node is bad only when it does not answer both on ipv4 and ipv6 addrs for N secods

src/toxcore/dht/dht_friend.rs
kurnevsky: This `if` is not about too many bootstrap nodes. It checks if we should request packet to a random node. Count of bootstrap nodes handles `Bucket` automatically here: https://github.com/tox-rs/tox/blob/master/src/toxcore/dht/dht_friend.rs#L150

kpp: Well it is better to describe what is going on then leave it uncommented

PR #177, opened 2018-07-20
Fix binary search

kurnevsky:


src/toxcore/dht/kbucket.rs
```diff
@@ -204,15 +202,27 @@ impl Bucket {
...
                 if self.is_full() {
-                    debug!(target: "Bucket",
-                        "Node is too distant to add to the bucket.");
-                    false
+                    match self.nodes.iter().rposition(|n| n.is_bad()) {
+                        Some(index) => {
+                            debug!(target: "Bucket",
+                                "No free space left in the bucket, the last bad node removed.");
+                            // replace the farthest bad node
+                            self.nodes.remove(index);
```
kpp: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove

PR #178, opened 2018-07-23
Tcp ping sender

NamsooCho:


src/toxcore/tcp/server/server.rs
```diff
@@ -1237,4 +1267,30 @@ mod tests {
             .wait();
         assert!(handle_res.is_err());
     }
+    #[test]
+    fn tcp_ping_sender_test() {
+        let server = Server::new();
+
+        // client #1
+        let (mut client_1, rx_1) = create_random_client();
+        client_1.set_last_pinged(Instant::now().sub(Duration::from_secs(TCP_PING_FREQUENCY + 1)));
```
kurnevsky: It won't always work - this sub can cause an integer overflow. See tokio_timer::clock, for example: https://github.com/tox-rs/tox/blob/3137d7a32d43a7a3971a6cda9dc0d40ef328cc11/src/toxcore/onion/onion_announce.rs#L565-L581

src/toxcore/tcp/server/client.rs
```diff
@@ -36,7 +38,11 @@ pub struct Client {
     */
     links: [Option<PublicKey>; 240],
     /// Used to check whether PongResponse is correct
-    ping_id: u64
+    ping_id: u64,
+    /// last time sent PingRequest packet
```
kurnevsky: Better start with capital letter to have uniform style.

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +373,25 @@ impl Server {
             Box::new( future::ok(()) )
         }
     }
+    /** Send periodic ping to TCP server
+    */
+    pub fn ping_sender(&self) -> IoFuture<()> {
```
kurnevsky: Better name it with verb, like `send_pings`.

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +373,25 @@ impl Server {
...
+        let mut state = self.state.write();
+
+        state.connected_clients.iter()
+            .filter(|(_key, client)| client.last_pong_resp().elapsed() >= Duration::from_secs(TCP_PING_FREQUENCY + TCP_PING_TIMEOUT))
```
kurnevsky: Better move this check to method of `Client`.

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +373,25 @@ impl Server {
...
+            .for_each(|(key, _client)| {
+                self.shutdown_client(key);
+            });
+
+        let ping_sender = state.connected_clients.iter_mut()
+            .filter(|(_key, client)| client.last_pinged().elapsed() >= Duration::from_secs(TCP_PING_FREQUENCY))
```
kurnevsky: Better move this check to method of Client.

src/toxcore/tcp/server/client.rs
```diff
@@ -3,11 +3,13 @@
 
 use toxcore::crypto_core::*;
 use toxcore::tcp::packet::*;
+use toxcore::tcp::packet::PingRequest;
```
kpp: I thought `use toxcore::tcp::packet::*;` includes all + PingRequest too

src/toxcore/tcp/server/server.rs
```diff
@@ -11,11 +11,17 @@ use std::io::{Error, ErrorKind};
...
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
+use std::time::{Instant, Duration};
 
 use futures::{Sink, Stream, Future, future, stream};
 use futures::sync::mpsc;
 use parking_lot::RwLock;
 
+/// Interval in seconds for sending TCP PingRequest
+pub const TCP_PING_FREQUENCY: u64 =  30;
```
kpp: 2 spaces in `=  30`

src/toxcore/tcp/server/client.rs
```diff
@@ -193,4 +221,20 @@ impl Client {
             Packet::Data(Data { connection_id, data })
         )
     }
+    /** Construct PingRequest and send it to Client
+    */
+    pub fn send_ping_request(&mut self) -> IoFuture<()> {
+        let mut ping_id = 0;
+
+        while(ping_id == 0) {
+            ping_id =  random_u64();
```
kpp: 2 spaces after `=`

NamsooCho: done

src/toxcore/tcp/server/client.rs
```diff
@@ -77,6 +90,38 @@ impl Client {
         self.ping_id
     }
 
+    /** Last time PingRequest sent.
+    */
+    pub fn last_pinged(&self) -> Instant { self.last_pinged }
+
+    /** Set last_pinged
+    */
+    pub fn set_last_pinged(&mut self, time: Instant) {
```
kurnevsky: No need in these two methods, since we have now more abstract `is_pong_timedout` and `is_ping_interval_passed`.

NamsooCho: done

kpp:
Also it would be nice if you integrated ping sender into tcp ServerExt

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +368,25 @@ impl Server {
             Box::new( future::ok(()) )
         }
     }
+    /** Send periodic ping to TCP server
```
kpp: `Send periodic ping to TCP server` -> `Send pings to all connected clients`

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +368,25 @@ impl Server {
...
     }
+    /** Send pings to all connected clients
+    */
+    pub fn send_pings(&self) -> IoFuture<()> {
+        let mut state = self.state.write();
+
+        state.connected_clients.iter()
+            .filter(|(_key, client)| client.is_pong_timedout())
+            .for_each(|(key, _client)| {
+                self.shutdown_client(key);
```
kurnevsky: `shutdown_client` returns future that should be handled.

src/toxcore/tcp/server/client.rs
```diff
@@ -193,4 +224,20 @@ impl Client {
             Packet::Data(Data { connection_id, data })
         )
     }
+    /** Construct PingRequest and send it to Client
+    */
+    pub fn send_ping_request(&mut self) -> IoFuture<()> {
+        let mut ping_id = 0;
+
+        while(ping_id == 0) {
```
kurnevsky: Missing space.
Btw, maybe add common function for generating non zero id?

src/toxcore/tcp/server/server.rs
```diff
@@ -1237,4 +1267,37 @@ mod tests {
             .wait();
         assert!(handle_res.is_err());
     }
+    #[test]
+    fn tcp_send_pings_test() {
```
kurnevsky: Add a test that ensures that timed out clients are removed.

src/toxcore/tcp/server/server.rs
```diff
@@ -1237,4 +1267,37 @@ mod tests {
...
+        let clock_1 = Clock::new_with_now(ConstNow(
+            now + Duration::from_secs(TCP_PING_FREQUENCY + 1)
+        ));
+
+        with_default(&clock_1, &mut enter, |_| {
+            let sender_res = server.send_pings().wait();
+            assert!(sender_res.is_ok());
+        });
+
+        let (_packet, _rx_1) = rx_1.into_future().wait().unwrap();
```
kurnevsky: Add check that `ping_id` from packets is the same that is stored inside clients.

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +368,25 @@ impl Server {
             Box::new( future::ok(()) )
         }
     }
+    /** Send pings to all connected clients
```
kurnevsky: Add note that this method removes timed out clients, something like this:
`Send pings to all connected clients and terminate all timed out clients.`

src/toxcore/tcp/server/server.rs
```diff
@@ -365,6 +368,41 @@ impl Server {
...
+    */
+    fn remove_timedout_clients(&self) -> IoFuture<()> {
+        let keys = self.state.read().connected_clients.iter()
+            .filter(|(_key, client)| client.is_pong_timedout())
+            .map(|(key, _client)| *key)
+            .collect::<Vec<PublicKey>>();
+
+        let remove_timedouts = keys.iter()
+            .map(|key| {
+                self.shutdown_client(key)
```
kurnevsky: Every call to this method acquires new lock which is not good. Besides performance impact it may lead to error when we try to remove already removed client (if it was removed after we collected keys). I'd prefer using single lock and creating inner `shutdown_client` function that takes state as a parameter instead of taking it from `RwLock` so that we can acquire this lock only once inside `send_pings`. What do you think?

NamsooCho: Good idea.

src/toxcore/tcp/server/server.rs
```diff
@@ -99,13 +100,20 @@ impl Server {
     */
     pub fn shutdown_client(&self, pk: &PublicKey) -> IoFuture<()> {
         let mut state = self.state.write();
+        let state = state.deref_mut();
```
kpp: Do we really need to use deref_mut op here?

NamsooCho: It just used for making type of state to simple.

NamsooCho: Ok I will remove it.

NamsooCho: done

src/toxcore/tcp/server/server.rs
```diff
@@ -98,13 +99,19 @@ impl Server {
...
+    fn shutdown_client_inner(&self, pk: &PublicKey, state: &mut ServerState) -> IoFuture<()> {
         let client_a = if let Some(client_a) = state.connected_clients.remove(pk) {
             client_a
         } else {
             return Box::new( future::err(
                 Error::new(ErrorKind::Other,
-                    "Cannot find client by pk to shutdown it"
-            )))
+                           "Cannot find client by pk to shutdown it"
+                )))
```
kurnevsky: Unnecessary diff.

src/toxcore/tcp/server/server.rs
```diff
@@ -127,7 +134,6 @@ impl Server {
             });
         Box::new( stream::futures_unordered(notifications).for_each(Ok) )
     }
-
```
kurnevsky: Unnecessary diff.

PR #179, opened 2018-07-24
DHT fixes

kurnevsky:


src/toxcore/dht/server/mod.rs
```diff
@@ -275,91 +310,61 @@ impl Server {
...
-
-            let mut random_node = random_usize() % good_nodes.len();
-            // increase probability of sending packet to a close node (has lower index)
-            if random_node != 0 {
-                random_node -= random_usize() % (random_node + 1);
-            }
-
-            let random_node = good_nodes[random_node];
+        if good_nodes.is_empty() {
+            return Box::new(future::ok(()))
```
kpp: plz add comment why we should return an empty future

src/toxcore/dht/server/mod.rs
```diff
@@ -275,91 +310,61 @@ impl Server {
...
+        Box::new(future::join_all(futures).map(|_| ()))
     }
 
     // every 60 seconds DHT node send ping(NodesRequest) to all nodes which is in close list
-    fn ping_and_get_close_nodes(&self) -> IoFuture<()> {
-        let mut close_nodes = self.close_nodes.write();
-        let mut request_queue = self.request_queue.write();
-
-        let nodes_sender = close_nodes.iter_mut()
+    fn ping_and_get_close_nodes<'a, T>(&self, request_queue: &mut RequestQueue, nodes: T, pk: PublicKey) -> IoFuture<()>
```
kpp: Does the name `ping_and_get_close_nodes` of the function reflect its behaviour?

kpp:
I can't bootstrap on this branch with `cargo run -- --udp-address 0.0.0.0:33445 --secret-key bf042bb227d0c7ff38302ead95a9e97e719182898e1cab8eab2ab6cda81de367 --no-lan -b 1D5A5F2F5D6233058BF0259B09622FB40B482E4FA0931EB8FD3AB8E7BF7DAF6F 198.98.51.198:33445`

kurnevsky:
@kpp you chose the wrong secret key for this bootstrap node :)
When we send `NodesRequest` packet to this bootstrap node we always receive the same nodes: ```74.199.52.131:33445, 188.116.134.156:33445, 65.52.199.26:33445, 217.244.81.144:33445```. And it seems all them are behind NAT so we can't reach them. So try to change you secret key or add another bootstrap node. This is expected behavior for now. Later we should implement fake friends feature so this won't be an issue.

src/toxcore/dht/server/mod.rs
```diff
@@ -573,26 +573,25 @@ impl Server {
...
-                return Box::new(future::err(Error::new(
-                    ErrorKind::Other,
-                    "DHT node is running in ipv4 mode but target node's socket is ipv6 address"
-                )))
+        } else {
+            if addr.is_ipv4() {
+                send_to(&self.tx, (packet, addr))
+            } else {
+                debug!("Attempting to send packet to IPv6 address wile DHT node is running in IPv4 mode");
+                Box::new(future::ok(()))
```
kpp: It's kinda strange since we may log errors in `tox-node` and convert them into `future::ok` only there. It will be better for 1 more reason: we can write tests

kpp:
> you chose the wrong secret key for this bootstrap node :)

This is a really good choice to find a PK that receives only NAT nodes =)

src/toxcore/dht/server/mod.rs
kurnevsky: Actually it's not exactly an error - we attempt sending IPv6 packets when receiving IPv6 nodes since we don't filter them in any way. This is the same behavior that c-toxcore has. And if we log error here it'll be confusing since it'll happen periodically in an absolutely normal run.

PR #181, opened 2018-07-30
Use DhtNode instead of PackedNode

NamsooCho:


src/toxcore/dht/dht_node.rs
```diff
@@ -31,9 +31,9 @@ If both node is Good node, then we compare PK's distance.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct DhtNode {
     /// Socket addr of node for IPv4.
-    pub saddr_v4: Option<SocketAddrV4>,
+    pub saddr_v4: Option<SocketAddr>,
     /// Socket addr of node for IPv6.
-    pub saddr_v6: Option<SocketAddrV6>,
+    pub saddr_v6: Option<SocketAddr>,
```
kurnevsky: Why change this to `SocketAddr`?

kurnevsky: Which means more opportunities to make a mistake. Isn't it wrong if `saddr_v4` stores IPv6 address or `saddr_v6` stores IPv4 address? I'd prefer to leave it as is but if you need these addresses as `SocketAddr` then just add two methods to `DhtNode` that converts them to `SocketAddr`.

NamsooCho: Ah. I got it. It is worse than before.

kurnevsky:
I think this `Into` implementation should be completely removed since it has incorrect in most cases logic and it's tempting to use it: https://github.com/tox-rs/tox/blob/168a5628047a236073af64a2153d5b7f0fc712f0/src/toxcore/dht/kbucket.rs#L52-L65

src/toxcore/dht/server/mod.rs
```diff
@@ -570,28 +569,33 @@ impl Server {
...
-            }
+    /// Send UDP packet node. If the node has both IPv4 and IPv6 addresses,
+    /// then it sends packet to both addresses.
+    fn send_to(&self, node: &DhtNode, packet: DhtPacket) -> IoFuture<()> {
+        let mut futures = Vec::new();
+
+        if self.is_ipv6_mode {// DHT node is running in ipv6 mode
+            if let Some(sock_v6) = node.saddr_v6 {
+                futures.push(send_to(&self.tx, (packet.clone(), SocketAddr::V6(sock_v6))));
+            } else {};
```
kurnevsky: Remove empty else block.

src/toxcore/dht/server/mod.rs
```diff
@@ -570,28 +569,33 @@ impl Server {
...
+
+            if let Some(sock_v4) = node.saddr_v4 {
+                let ip_v6 = sock_v4.ip().to_ipv6_mapped();
+                let addr = SocketAddr::new(IpAddr::V6(ip_v6), sock_v4.port());
+                futures.push(send_to(&self.tx, (packet, addr)));
+            } else {};
```
kurnevsky: Remove empty else block.

src/toxcore/dht/server/mod.rs
```diff
@@ -570,28 +569,33 @@ impl Server {
...
+        } else { // DHT node is running in ipv4 mode
+            if let Some(sock_v4) = node.saddr_v4 {
+                futures.push(send_to(&self.tx, (packet, SocketAddr::V4(sock_v4))));
+            } else {};
```
kurnevsky: Remove empty else block.

src/toxcore/dht/server/mod.rs
```diff
@@ -570,28 +569,33 @@ impl Server {
...
+            if let Some(sock_v4) = node.saddr_v4 {
+                futures.push(send_to(&self.tx, (packet, SocketAddr::V4(sock_v4))));
+            };
         }
+
+        Box::new(join_all(futures).map(|_| ()))
+    }
+
+    /// Send UDP packet to specified address.
+    fn send_to_resp(&self, addr: SocketAddr, packet: DhtPacket) -> IoFuture<()> {
```
kpp: rename to `send_to_direct`

src/toxcore/dht/server/mod.rs
```diff
@@ -570,28 +569,33 @@ impl Server {
...
-        } else {
-            if addr.is_ipv4() {
-                send_to(&self.tx, (packet, addr))
-            } else {
-                debug!("Attempting to send packet to IPv6 address wile DHT node is running in IPv4 mode");
-                Box::new(future::ok(()))
-            }
+    /// Send UDP packet node. If the node has both IPv4 and IPv6 addresses,
+    /// then it sends packet to both addresses.
+    fn send_to(&self, node: &DhtNode, packet: DhtPacket) -> IoFuture<()> {
```
kpp: rename to `send_to_node`

src/toxcore/dht/server/mod.rs
```diff
@@ -570,28 +569,33 @@ impl Server {
...
+                let ip_v6 = sock_v4.ip().to_ipv6_mapped();
+                let addr = SocketAddr::new(IpAddr::V6(ip_v6), sock_v4.port());
+                futures.push(send_to(&self.tx, (packet, addr)));
+            };
+        } else { // DHT node is running in ipv4 mode
+            if let Some(sock_v4) = node.saddr_v4 {
+                futures.push(send_to(&self.tx, (packet, SocketAddr::V4(sock_v4))));
+            };
         }
+
```
kpp: Add warning if `futures` are empty, something like: `Trying to send a packet to node PK but could not find appropriate addr`

kpp:
> I think this Into implementation should be completely removed since it has incorrect in most cases logic and it's tempting to use it: 

I agree

examples/dht_server.rs
```diff
@@ -75,7 +75,7 @@ fn main() {
 
     let lan_discovery_sender = LanDiscoverySender::new(tx.clone(), server_pk, local_addr.is_ipv6());
 
-    let mut server = Server::new(tx, server_pk, server_sk);
+    let mut server = Server::new(tx, server_pk, server_sk, local_addr.is_ipv6());
```
kpp: `server.set_ipv6_mode` looks better.

src/toxcore/dht/kbucket.rs
```diff
@@ -115,6 +100,8 @@ pub struct Bucket {
     pub capacity: u8,
     /// Nodes that bucket has, sorted by distance to PK.
     pub nodes: Vec<DhtNode>,
+    /// Flag for ipv6 enabled or not
+    pub is_ipv6_enabled: bool,
```
kpp: Did you use this flag anywhere?

src/toxcore/dht/dht_node.rs
```diff
@@ -123,6 +123,37 @@ impl DhtNode {
             }
         }
     }
+
+    /// convert Dhtnode to PackedNode object based on is_ipv6_enabled flag
+    pub fn to_packed_node(self, is_ipv6_enabled: bool) -> PackedNode {
+        let saddr = if is_ipv6_enabled {
+            match self.saddr_v6 {
```
kurnevsky: Can we unify this with `get_socket_addr`? Perhaps make `get_socket_addr` return address without `Option`?

kpp: > Or make this method to return Option.

sounds better

examples/dht_server.rs
NamsooCho: set_ipv6_mode() should set Server, DhtFriend, Bucket, Kbucket. This increases complexity of source code.
And there is no benefit for dynamic setting for ip_v6, I think.
Does it need to set ip_v6 dynamically?

src/toxcore/dht/kbucket.rs
NamsooCho: https://github.com/tox-rs/tox/blob/7cdd81e5605f3e1f3c6d386babfdf99728e3480b/src/toxcore/dht/kbucket.rs#L306

kpp: You may pass it as arg like you did before in `.to_packed_node(self.is_ipv6_enabled))`

examples/dht_server.rs
kurnevsky: It's not necessary that `Bucket` contains `is_ipv6_enabled`. If you don't add to `Bucket` it then `set_ipv6_mode` will have to update only `Server` and `Kbucket` (and perhaps we will be able to remove it from `Kbucket` later as well).

src/toxcore/dht/dht_node.rs
```diff
@@ -123,6 +123,20 @@ impl DhtNode {
...
+
+    /// convert Dhtnode to PackedNode object based on is_ipv6_enabled flag
+    pub fn to_packed_node(self, is_ipv6_enabled: bool) -> Option<PackedNode> {
+        if let Some(saddr) = self.get_socket_addr(is_ipv6_enabled) {
+            Some(PackedNode {
+                pk: self.pk,
+                saddr,
+            })
+        } else {
+            warn!("to_packed_node() fails because get_socket_addr returns None");
```
kurnevsky: I think this warning can be moved to `get_socket_addr` since it should never return `None`. And if you do this then this `if` can be converted to simple `map`.

NamsooCho: done

src/toxcore/dht/dht_node.rs
```diff
@@ -112,17 +112,34 @@ impl DhtNode {
                 None => {
                     match self.saddr_v4 {
                         Some(v4) => Some(SocketAddr::V4(v4)),
-                        None => None,
+                        None => {
+                            warn!("to_packed_node() fails because get_socket_addr returns None");
```
kurnevsky: Correct message text to something like:
"get_socket_addr: failed to get address of DhtNode"

src/toxcore/dht/dht_node.rs
```diff
@@ -112,17 +112,34 @@ impl DhtNode {
...
+                        },
                     }
                 },
             }
         } else {
             match self.saddr_v4 {
                 Some(v4) => Some(SocketAddr::V4(v4)),
-                None => None,
+                None => {
+                    warn!("to_packed_node() fails because get_socket_addr returns None");
```
kurnevsky: Correct message text to something like:
"get_socket_addr: failed to get address of DhtNode"

src/toxcore/dht/dht_node.rs
```diff
@@ -94,8 +94,8 @@ impl DhtNode {
     }
 
     /// return SocketAddr for DhtNode
-    pub fn get_socket_addr(&self, is_ipv6_mode: bool) -> Option<SocketAddr> {
-        if is_ipv6_mode {
+    pub fn get_socket_addr(&self, is_ipv6_enabled: bool) -> Option<SocketAddr> {
+        if is_ipv6_enabled {
```
kpp: See logic https://github.com/tox-rs/tox/pull/183/files#diff-77f898031e13d02915c0d710043ed97bR142 

Please don't duplicate code. It is better to do PRs 1 by 1

src/toxcore/dht/server/mod.rs
```diff
@@ -569,28 +567,35 @@ impl Server {
...
+                futures.push(send_to(&self.tx, (packet.clone(), SocketAddr::V6(sock_v6))));
+            };
+
+            if let Some(sock_v4) = node.assoc4.saddr {
+                let ip_v6 = sock_v4.ip().to_ipv6_mapped();
+                let addr = SocketAddr::new(IpAddr::V6(ip_v6), sock_v4.port());
+                futures.push(send_to(&self.tx, (packet, addr)));
+            };
+        } else if let Some(sock_v4) = node.assoc4.saddr { // DHT node is running in ipv4 mode
+                futures.push(send_to(&self.tx, (packet, SocketAddr::V4(sock_v4))));
```
kurnevsky: Wrong padding.

src/toxcore/dht/server/mod.rs
```diff
@@ -432,7 +432,7 @@ impl Server {
             payload
         ));
 
-        self.send_to(target_peer.saddr, nodes_req)
+        self.send_to_node(target_peer, nodes_req)
```
kurnevsky: c-toxcore sends two packets with different ping ids so it will process both responses. If we send same packet we will be able to process only one response.

NamsooCho: NodesRequest sends packets with different ping_id to both IPv4 and IPv6.
PingRequest is not sending packet to both addresses.
NatPingRequest uses same ping_id to both addresses.

src/toxcore/dht/server/mod.rs
```diff
@@ -132,7 +133,7 @@ pub struct Server {
     lan_discovery_enabled: bool,
     /// If IPv6 mode is enabled `Server` will send packets to IPv6 addresses. If
     /// it's disabled such packets will be dropped.
-    is_ipv6_mode: bool,
+    pub is_ipv6_enabled: bool,
```
kpp: why pub?

src/toxcore/dht/dht_node.rs
```diff
@@ -155,14 +155,36 @@ impl DhtNode {
                     }
                 },
                 None => {
-                    self.assoc4.saddr.map(Into::into)
+                    match self.assoc4.saddr {
```
kpp: Did you rebase successfully?

src/toxcore/dht/server/mod.rs
NamsooCho: https://github.com/tox-rs/tox/blob/24cfbb1821fa9e71e4832544e3011f14c6fe8f73/src/toxcore/dht/daemon_state.rs#L40

src/toxcore/dht/dht_node.rs
NamsooCho: It is changed to leave warning because if it is None then something goes wrong.

NamsooCho: https://github.com/tox-rs/tox/blob/24cfbb1821fa9e71e4832544e3011f14c6fe8f73/src/toxcore/dht/dht_node.rs#L170-L173

src/toxcore/dht/server/mod.rs
kpp: well you may add getter for this field.

src/toxcore/dht/dht_node.rs
kpp: https://doc.rust-lang.org/std/option/enum.Option.html#method.or_else

src/toxcore/dht/dht_node.rs
```diff
@@ -154,15 +154,31 @@ impl DhtNode {
                         None => Some(SocketAddr::V6(v6)),
                     }
                 },
-                None => {
-                    self.assoc4.saddr.map(Into::into)
-                },
+                None => self.assoc4.saddr.map(Into::into)
+                    .or_else(|| {
```
kurnevsky: We can just set variable `let result = ...` before returning the address. And then we can warn only if it's empty. This will reduce two checks/calls to one.

src/toxcore/dht/server/mod.rs
```diff
@@ -158,14 +159,19 @@ impl Server {
...
     /// Enable/disable IPv6 mode of DHT server.
     pub fn enable_ipv6_mode(&mut self, enable: bool) {
-        self.is_ipv6_mode = enable;
-        self.close_nodes.write().is_ipv6_mode = enable;
+        self.is_ipv6_enabled = enable;
+        self.close_nodes.write().is_ipv6_enabled = enable;
+    }
+
+    /// Get is_ipv6_enabled member variable
+    pub fn get_is_ipv6_enabled(&self) -> bool {
```
kpp: just `is_ipv6_enabled` should be OK

src/toxcore/dht/server/mod.rs
```diff
@@ -408,11 +415,11 @@ impl Server {
             &self.pk,
             payload
         ));
-        self.send_to(node.saddr, ping_req)
+        self.send_to_direct(node.saddr, ping_req)
     }
 
     /// Send `NodesRequest` packet to the node.
-    pub fn send_nodes_req(&self, target_peer: PackedNode, search_pk: PublicKey, ping_id: u64) -> IoFuture<()> {
+    pub fn send_nodes_req(&self, target_peer: &DhtNode, search_pk: PublicKey, request_queue: &mut RequestQueue) -> IoFuture<()> {
```
kpp: All methods accepting `request_queue` keep this arg right after `self`.

src/toxcore/dht/server/mod.rs
```diff
@@ -422,17 +429,43 @@ impl Server {
             )
         }
 
+        let mut futures = Vec::new();
+
+        if self.is_ipv6_enabled {
+            if let Some(v6) = target_peer.assoc6.saddr {
```
kpp: Try to use https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then instead of `if let Some`

src/toxcore/dht/server/mod.rs
```diff
@@ -422,17 +429,43 @@ impl Server {
             )
         }
 
+        let mut futures = Vec::new();
+
+        if self.is_ipv6_enabled {
```
kurnevsky: Maybe add method to `DhtNode` that takes `is_ipv6_enabled` and returns `Vec<SocketAddr>`? Then we won't need to duplicate this logic here and in `send_to_node` method.

src/toxcore/dht/server/mod.rs
```diff
@@ -755,13 +794,11 @@ impl Server {
             }
         }
 
-        friends.iter_mut()
-            .find(|friend| friend.pk == node.pk)
-            .map(|friend| {
```
kpp: I prefer `.map` more than `if let Some`

src/toxcore/dht/dht_node.rs
kpp: Or we can check it right where we call this function

src/toxcore/dht/server/mod.rs
```diff
@@ -422,17 +429,36 @@ impl Server {
             )
         }
 
+        let mut futures = Vec::new();
+
+        let addrs = target_peer.get_all_sockets(self.is_ipv6_enabled);
+
+        addrs.into_iter()
```
kurnevsky: Better to use `map` here, like `let futures = addrs.map(...).collect()`.

src/toxcore/dht/server/mod.rs
```diff
@@ -569,28 +593,28 @@ impl Server {
...
-                Box::new(future::ok(()))
-            }
+    /// Send UDP packet node. If the node has both IPv4 and IPv6 addresses,
+    /// then it sends packet to both addresses.
+    fn send_to_node(&self, node: &DhtNode, packet: DhtPacket) -> IoFuture<()> {
+        let mut futures = Vec::new();
+
+        let addrs = node.get_all_sockets(self.is_ipv6_enabled);
+
+        addrs.into_iter()
```
kurnevsky: Better to use `map` here, like `let futures = addrs.map(...).collect()`.

src/toxcore/dht/dht_node.rs
```diff
@@ -154,13 +154,44 @@ impl DhtNode {
...
+
+        addr
+    }
+
+    /// Returns all available socket addresses of DhtNode
+    pub fn get_all_sockets(&self, is_ipv6_enabled: bool) -> Vec<SocketAddr> {
+        let mut addrs = Vec::new();
+
+        if is_ipv6_enabled {// DHT node is running in ipv6 mode
+            self.assoc6.saddr.and_then(|v6| Some(addrs.push(SocketAddr::V6(v6))));
```
kurnevsky: `if let` is better here since you don't care of the result of `and_then` and mutate `addrs` inside.

src/toxcore/dht/dht_node.rs
```diff
@@ -154,13 +154,44 @@ impl DhtNode {
...
+            self.assoc4.saddr.and_then(|v4| {
+                let ip_v6 = v4.ip().to_ipv6_mapped();
+                Some(addrs.push(SocketAddr::new(IpAddr::V6(ip_v6), v4.port())))
+            });
+        } else { // DHT node is running in ipv4 mode
+            self.assoc4.saddr.and_then(|v4| Some(addrs.push(SocketAddr::V4(v4))));
```
kurnevsky: `if let` is better here since you don't care of the result of `and_then` and mutate `addrs` inside.

src/toxcore/dht/dht_node.rs
```diff
@@ -154,13 +154,44 @@ impl DhtNode {
...
+        addr
+    }
+
+    /// Returns all available socket addresses of DhtNode
+    pub fn get_all_sockets(&self, is_ipv6_enabled: bool) -> Vec<SocketAddr> {
+        let mut addrs = Vec::new();
+
+        if is_ipv6_enabled {// DHT node is running in ipv6 mode
+            self.assoc6.saddr.and_then(|v6| Some(addrs.push(SocketAddr::V6(v6))));
+            self.assoc4.saddr.and_then(|v4| {
```
kurnevsky: `if let` is better here since you don't care of the result of `and_then` and mutate `addrs` inside.

src/toxcore/dht/server/mod.rs
```diff
@@ -569,28 +593,28 @@ impl Server {
...
+            .for_each(|addr| {
+                futures.push(send_to(&self.tx, (packet.clone(), addr)));
+            });
+
+        if futures.is_empty() {
```
kurnevsky: Now we can move this warning to `get_all_sockets` so that any usage of this method will make it.

src/toxcore/dht/server/mod.rs
```diff
@@ -422,17 +429,36 @@ impl Server {
...
+
+        let addrs = target_peer.get_all_sockets(self.is_ipv6_enabled);
+
+        addrs.into_iter()
+            .for_each(|addr| {
+                let ping_id = request_queue.new_ping_id(target_peer.pk);
+                futures.push(self.send_nodes_req_inner(addr, target_peer.pk, search_pk, ping_id));
+            });
+
+        if futures.is_empty() {
```
kurnevsky: Now we can move this warning to `get_all_sockets` so that any usage of this method will make it.

src/toxcore/dht/dht_node.rs
```diff
@@ -154,13 +154,44 @@ impl DhtNode {
...
+
+        if addr.is_none() {
+            warn!("get_socket_addr: failed to get address of DhtNode");
         }
+
+        addr
+    }
+
+    /// Returns all available socket addresses of DhtNode
+    pub fn get_all_sockets(&self, is_ipv6_enabled: bool) -> Vec<SocketAddr> {
```
kurnevsky: `get_all_addrs`?

src/toxcore/dht/dht_node.rs
kpp: why?

kurnevsky: See this line:
```rust
self.assoc4.saddr.and_then(|v4| Some(addrs.push(SocketAddr::V4(v4))));
```
First, we ignore the result of `and_then` which returns `Option` itself.
Second, we artificially pack `addrs.push(SocketAddr::V4(v4))` in `Some` inside so we get `Option<()>` which is kind of ugly.
Third, all we do inside is mutating `addrs` variable so this `and_then` is not clean.

Aren't enough reasons to use simple `if let` construction? :)

kurnevsky: If you want it in functional style then it should be like:
```rust
self.assoc4.iter().map(SocketAddr::V4)
    .chain(self.assoc6.iter().map(SocketAddr::V6))
    .collect()
```
But not the current approach.

kpp: OK, I agree

src/toxcore/dht/server/mod.rs
kurnevsky: https://stackoverflow.com/a/27535000

kpp: Thanks

src/toxcore/dht/server/mod.rs
```diff
@@ -750,18 +765,14 @@ impl Server {
     /// Update returned socket address and time of receiving packet
     fn update_returned_addr(&self, node: &PackedNode, packet_pk: &PublicKey, close_nodes: &mut Kbucket, friends: &mut Vec<DhtFriend>) {
         if self.pk == node.pk {
-            if let Some(node_to_update) = close_nodes.get_node_mut(packet_pk) {
-                node_to_update.update_returned_addr(node.saddr);
-            }
+            close_nodes.get_node_mut(packet_pk)
+                .and_then(|node_to_update| Some(node_to_update.update_returned_addr(node.saddr)));
```
kpp: Revert to `if let`

src/toxcore/dht/server/mod.rs
```diff
@@ -750,18 +765,14 @@ impl Server {
...
-        friends.iter_mut()
-            .find(|friend| friend.pk == node.pk)
-            .map(|friend| {
-                if let Some(node_to_update) = friend.close_nodes.get_node_mut(&friend.pk, packet_pk) {
-                    node_to_update.update_returned_addr(node.saddr);
-                }
-            });
+        if let Some(friend) = friends.iter_mut().find(|friend| friend.pk == node.pk) {
+            friend.close_nodes.get_node_mut(&friend.pk, packet_pk)
+                .and_then(|node_to_update| Some(node_to_update.update_returned_addr(node.saddr)));
```
kpp: Revert to `if let`

PR #182, opened 2018-07-31
Simplify example and synchronize it with tox-node

kurnevsky:


examples/dht_server.rs
```diff
@@ -15,152 +15,89 @@ use futures::sync::mpsc;
...
     env_logger::init();
 
     if !crypto_init() {
-        error!("Crypto initialization failed.");
-        return;
+        panic!("Crypto initialization failed.");
     }
 
-    let first_arg = env::args().nth(1);
-    let (server_pk, server_sk) = if first_arg == Some("--const_key".to_string()) {
```
kpp: How about --const_key?

PR #183, opened 2018-08-01
Hole punch: take address of friend they returned

NamsooCho:
https://zerhoinfo.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=SL&modal=detail&selectedIssue=SL-237

I think SL - 237 is for this PR

### DIFF
--- a/src/toxcore/dht/dht_friend.rs
+++ b/src/toxcore/dht/dht_friend.rs
@@ -9,6 +9,7 @@ use toxcore::time::*;
 use toxcore::dht::kbucket::*;
 use toxcore::crypto_core::*;
 use toxcore::dht::server::hole_punching::*;
+use toxcore::dht::dht_node::*;
 
 /// Number of close nodes each friend has.
 pub const FRIEND_CLOSE_NODES_COUNT: u8 = 4;
@@ -58,33 +59,31 @@ impl DhtFriend {
 
     /// get Socket Address list of a friend, a friend can have multi IP address bacause of NAT
     pub fn get_addrs_of_clients(&self) -> Vec<SocketAddr> {
-        let mut socks = Vec::new();
-        let mut direct_connected = false;
+        let mut addrs = Vec::new();
 
-        self.close_nodes.nodes.iter()
-            .for_each(|node| {
-                if let Some(v6) = node.assoc6.ret_saddr {
-                    if !node.assoc6.is_bad() {
-                        socks.push(v6);
-                    }
+        let good_nodes = self.close_nodes.nodes.iter()
+            .filter(|node| !node.is_bad())
+            .cloned()
+            .collect::<Vec<DhtNode>>();
+
+        for node in good_nodes {
+            if let Some(v6) = node.assoc6.ret_saddr {
+                if !node.assoc6.is_bad() {
+                    addrs.push(SocketAddr::V6(v6));
                 }
+            }
 
-                if let Some(v4) = node.assoc4.ret_saddr {
-                    if !node.assoc4.is_bad() {
-                        socks.push(v4);
-                    }
+            if let Some(v4) = node.assoc4.ret_saddr {
+                if !node.assoc4.is_bad() {
+                    addrs.push(SocketAddr::V4(v4));
                 }
+            }
 
-                if self.pk == node.pk && (!node.assoc4.is_bad() || !node.assoc6.is_bad()) {
-                    direct_connected = true;
-                }
-            });
-
-        if direct_connected {
-            Vec::new()
-        } else {
-            socks
+            if self.pk == node.pk {
+                return Vec::new();
+            }
         }
+        addrs
     }
 }
 
--- a/src/toxcore/dht/dht_friend.rs
+++ b/src/toxcore/dht/dht_friend.rs
@@ -58,21 +58,20 @@ impl DhtFriend {
 
     /// get Socket Address list of a friend, a friend can have multi IP address bacause of NAT
     pub fn get_addrs_of_clients(&self) -> Vec<SocketAddr> {
-        let mut sock_v4 = Vec::new();
-        let mut sock_v6 = Vec::new();
+        let mut socks = Vec::new();
         let mut direct_connected = false;
 
         self.close_nodes.nodes.iter()
             .for_each(|node| {
                 if let Some(v6) = node.assoc6.ret_saddr {
                     if !node.assoc6.is_bad() {
-                        sock_v6.push(v6);
+                        socks.push(v6);
                     }
                 }
 
                 if let Some(v4) = node.assoc4.ret_saddr {
                     if !node.assoc4.is_bad() {
-                        sock_v4.push(v4);
+                        socks.push(v4);
                     }
                 }
 
@@ -83,10 +82,8 @@ impl DhtFriend {
 
         if direct_connected {
             Vec::new()
-        } else if sock_v6.len() > sock_v4.len() {
-            sock_v6
         } else {
-            sock_v4
+            socks
         }
     }
 }
--- a/src/toxcore/dht/dht_friend.rs
+++ b/src/toxcore/dht/dht_friend.rs
@@ -57,11 +57,37 @@ impl DhtFriend {
     }
 
     /// get Socket Address list of a friend, a friend can have multi IP address bacause of NAT
-    pub fn get_addrs_of_clients(&self, is_ipv6_mode: bool) -> Vec<SocketAddr> {
+    pub fn get_addrs_of_clients(&self) -> Vec<SocketAddr> {
+        let mut sock_v4 = Vec::new();
+        let mut sock_v6 = Vec::new();
+        let mut direct_connected = false;
+
         self.close_nodes.nodes.iter()
-            .map(|node| node.get_socket_addr(is_ipv6_mode))
-            .filter_map(|addr| addr)
-            .collect::<Vec<SocketAddr>>()
+            .for_each(|node| {
+                if let Some(v6) = node.assoc6.ret_saddr {
+                    if !node.assoc6.is_bad() {
+                        sock_v6.push(v6);
+                    }
+                }
+
+                if let Some(v4) = node.assoc4.ret_saddr {
+                    if !node.assoc4.is_bad() {
+                        sock_v4.push(v4);
+                    }
+                }
+
+                if self.pk == node.pk && (!node.assoc4.is_bad() || !node.assoc6.is_bad()) {
+                    direct_connected = true;
+                }
+            });
+
+        if direct_connected {
+            Vec::new()
+        } else if sock_v6.len() > sock_v4.len() {
+            sock_v6
+        } else {
+            sock_v4
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


kurnevsky:
Still wrong. You mixed up IP addresses of close nodes to our friend with IP addresses of this friend that they return. See the difference between `ip_port` and `ret_ip_port`: https://github.com/TokTok/c-toxcore/blob/3a4987da18b165d5a5297ddaf64f1336958bb3c7/toxcore/DHT.c#L1911-L1919

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,24 @@ impl Server {
         }
     }
 
+    /// Update returned socket address and time of receiving packet
+    fn update_returned_sock(&self, node: &PackedNode, close_nodes: &mut Kbucket, friends: &mut Vec<DhtFriend>) {
+        if self.pk == node.pk {
```
kurnevsky: Node with the same pk as ours can't be stored in kbucket. So this if is no-op.

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,24 @@ impl Server {
...
+            if let Some(node_to_update) = close_nodes.get_node_mut(&node.pk) {
+                node_to_update.update_returned_sock(node.saddr);
+            } else {}
+        }
+
+        friends.iter_mut()
+            .for_each(|friend| {
+                if friend.pk == node.pk {
+                    if let Some(node_to_update) = friend.close_nodes.get_node_mut(&friend.pk, &node.pk) {
```
kurnevsky: Wrong pk is used. You should update node that sent `NodesResponse`. `node.pk` should be friend's pk.

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,24 @@ impl Server {
...
+                        node_to_update.update_returned_sock(node.saddr);
+                    } else {}
+                } else {}
```
kurnevsky: No need in such else blocks.

src/toxcore/dht/dht_friend.rs
```diff
@@ -57,54 +57,46 @@ impl DhtFriend {
     }
 
     /// get Socket Address list of a friend, a friend can have multi IP address bacause of NAT
-    pub fn get_addrs_of_clients(&self, is_ipv6_mode: bool) -> Vec<SocketAddr> {
+    pub fn get_addrs_of_clients(&self) -> Vec<SocketAddr> {
+        let mut sock_v4 = Vec::new();
```
kurnevsky: It seems we don't need to split IPv4 and IPv6 addresses here. `c-toxcore` does this because of flaky implementation.

src/toxcore/dht/dht_node.rs
```diff
@@ -30,30 +53,24 @@ If both node is Good node, then we compare PK's distance.
 */
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct DhtNode {
-    /// Socket addr of node for IPv4.
-    pub saddr_v4: Option<SocketAddrV4>,
-    /// Socket addr of node for IPv6.
-    pub saddr_v6: Option<SocketAddrV6>,
+    /// Socket addr and times of node for IPv4.
+    pub assoc4: SockAndTime,
```
kpp: Again, `pub saddr: Option<SocketAddr>` from SockAndTime may keep IPv6 (from the point of view of the type system).

src/toxcore/dht/dht_node.rs
```diff
@@ -22,6 +22,52 @@ pub const BAD_NODE_TIMEOUT: u64 = PING_INTERVAL * 2 + 2;
 /// The timeout after which a node is discarded completely.
 pub const KILL_NODE_TIMEOUT: u64 = BAD_NODE_TIMEOUT + PING_INTERVAL;
 
+/// Struct conatains SocketAddrs for IPv4 and timestamps for sending and receiving packet
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SockAndTimeV4 {
```
kpp: How about generic types?

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,24 @@ impl Server {
         }
     }
 
+    /// Update returned socket address and time of receiving packet
+    fn update_returned_sock(&self, node: &PackedNode, packet_pk: &PublicKey, close_nodes: &mut Kbucket, friends: &mut Vec<DhtFriend>) {
+        if self.pk == node.pk {
```
kurnevsky: We don't need this `if`. It will always do nothing.

src/toxcore/dht/dht_friend.rs
```diff
@@ -57,54 +57,43 @@ impl DhtFriend {
     }
 
     /// get Socket Address list of a friend, a friend can have multi IP address bacause of NAT
-    pub fn get_addrs_of_clients(&self, is_ipv6_mode: bool) -> Vec<SocketAddr> {
+    pub fn get_addrs_of_clients(&self) -> Vec<SocketAddr> {
+        let mut socks = Vec::new();
```
kurnevsky: socks -> addrs

src/toxcore/dht/dht_friend.rs
```diff
@@ -57,54 +57,43 @@ impl DhtFriend {
...
+                }
+
+                if let Some(v4) = node.assoc4.ret_saddr {
+                    if !node.assoc4.is_bad() {
+                        socks.push(SocketAddr::V4(v4));
+                    }
+                }
+
+                if self.pk == node.pk && (!node.assoc4.is_bad() || !node.assoc6.is_bad()) {
+                    direct_connected = true;
```
kurnevsky: No need in this variable. You can return empty vector right here.

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,24 @@ impl Server {
...
+            if let Some(node_to_update) = close_nodes.get_node_mut(packet_pk) {
+                node_to_update.update_returned_sock(node.saddr);
+            }
+        }
+
+        friends.iter_mut()
+            .for_each(|friend| {
+                if friend.pk == node.pk {
```
kurnevsky: Use `Iterator::find`.

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,24 @@ impl Server {
         }
     }
 
+    /// Update returned socket address and time of receiving packet
+    fn update_returned_sock(&self, node: &PackedNode, packet_pk: &PublicKey, close_nodes: &mut Kbucket, friends: &mut Vec<DhtFriend>) {
```
kurnevsky: `update_returned_sock` -> `update_returned_addr` since we update address and network socket has different meaning.

src/toxcore/dht/dht_node.rs
kpp: You may merge SockAndTimeV4 and SockAndTimeV6 using generics

src/toxcore/dht/server/mod.rs
NamsooCho: A node may have my DHT pk in it's Kbucket. So, my DHT pk may exist in NodesResponse.
Then the node's PK can be updated in my ret_saddr. Is it wrong?

src/toxcore/dht/dht_friend.rs
NamsooCho: Is it possible to break for_each()?

kpp: Sure.

src/toxcore/dht/dht_node.rs
```diff
@@ -64,43 +81,68 @@ impl DhtNode {
...
     pub fn is_bad(&self) -> bool {
-        self.last_resp_time_v4.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(BAD_NODE_TIMEOUT)) &&
-            self.last_resp_time_v6.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(BAD_NODE_TIMEOUT))
+        self.is_ipv4_bad() && self.is_ipv6_bad()
+    }
+
+    /// Check if the IPv4 address is considered bad i.e. it does not answer on IPv4
+    /// addresses for `BAD_NODE_TIMEOUT` seconds.
+    fn is_ipv4_bad(&self) -> bool {
+        self.assoc4.last_resp_time.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(BAD_NODE_TIMEOUT))
```
kpp: reuse code : `self.assoc4.is_bad()`

src/toxcore/dht/dht_node.rs
```diff
@@ -64,43 +81,68 @@ impl DhtNode {
...
+    fn is_ipv6_bad(&self) -> bool {
+        self.assoc6.last_resp_time.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(BAD_NODE_TIMEOUT))
     }
 
     /// Check if the node is considered discarded i.e. it does not answer both
     /// on IPv4 and IPv6 addresses for `KILL_NODE_TIMEOUT` seconds.
     pub fn is_discarded(&self) -> bool {
-        self.last_resp_time_v4.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(KILL_NODE_TIMEOUT)) &&
-            self.last_resp_time_v6.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(KILL_NODE_TIMEOUT))
+        self.assoc4.last_resp_time.map_or(true, |time| clock_elapsed(time) > Duration::from_secs(KILL_NODE_TIMEOUT)) &&
```
kpp: add method SockAndTime::is_discarded and use it here

NamsooCho: done

src/toxcore/dht/dht_friend.rs
```diff
@@ -57,54 +58,41 @@ impl DhtFriend {
...
-    pub fn get_addrs_of_clients(&self, is_ipv6_mode: bool) -> Vec<SocketAddr> {
-        self.close_nodes.nodes.iter()
-            .map(|node| node.get_socket_addr(is_ipv6_mode))
-            .filter_map(|addr| addr)
-            .collect::<Vec<SocketAddr>>()
+    pub fn get_addrs_of_clients(&self) -> Vec<SocketAddr> {
+        let mut addrs = Vec::new();
+
+        let good_nodes = self.close_nodes.nodes.iter()
+            .filter(|node| !node.assoc4.is_bad() || !node.assoc6.is_bad())
```
kpp: I guess it should be `!node.assoc4.is_bad() && !node.assoc6.is_bad()`

src/toxcore/dht/dht_node.rs
```diff
@@ -110,19 +161,43 @@ impl DhtNode {
                     }
                 },
                 None => {
-                    match self.saddr_v4 {
+                    match self.assoc4.saddr {
```
kpp:
```
                    match self.assoc4.saddr {
                        Some(v4) => Some(SocketAddr::V4(v4)),
                        None => None,
                    }
```

I think this code can be reduced into a smaller one. Assume you have `Option<SocketAddrV4>` in `self.assoc4.sadd` and you need to cast it to `Option<SocketAddr>`. You may apply `self.assoc4.saddr.map(Into::into)` which should cast `SocketAddrV4` to `SocketAddr` inside `Option`

src/toxcore/dht/dht_friend.rs
kurnevsky: It seems correct - node is good when we received packet from at least one address, not necessary both. But anyway `DhtNode` has method `is_bad` which can be used here.

src/toxcore/dht/dht_node.rs
```diff
@@ -64,43 +92,66 @@ impl DhtNode {
             (None, Some(clock_now()))
         };
 
+        let assoc4 = SockAndTime {
+            saddr: saddr_v4,
+            last_resp_time: last_resp_time_v4,
+            last_ping_req_time: None,
```
kurnevsky: You can create `new` method to omit all these `None` values.

src/toxcore/dht/dht_friend.rs
```diff
@@ -57,54 +58,41 @@ impl DhtFriend {
...
+                }
+            }
+
+            if let Some(v4) = node.assoc4.ret_saddr {
+                if !node.assoc4.is_bad() {
+                    addrs.push(SocketAddr::V4(v4));
+                }
+            }
+
+            if self.pk == node.pk && (!node.assoc4.is_bad() || !node.assoc6.is_bad()) {
```
kurnevsky: `&& !node.is_bad()`

src/toxcore/dht/server/mod.rs
```diff
@@ -746,6 +748,23 @@ impl Server {
         }
     }
 
+    /// Update returned socket address and time of receiving packet
+    fn update_returned_addr(&self, node: &PackedNode, packet_pk: &PublicKey, close_nodes: &mut Kbucket, friends: &mut Vec<DhtFriend>) {
+        if self.pk == node.pk {
```
kurnevsky: > A node may have my DHT pk in it's Kbucket. So, my DHT pk may exist in NodesResponse.
Then the node's PK can be updated in my ret_saddr. Is it wrong?

Right, NodesResponse can have our DHT pk but we can't have ourself in our kbucket. See: https://github.com/tox-rs/tox/blob/776a07f6e4d466fec562b1f1c879452baf4684b7/src/toxcore/dht/kbucket.rs#L28-L50

So this if is useless and can be removed.

NamsooCho: we search Kbucket using packet_pk.

kurnevsky: ok, you are right. I looked at the wrong key.

src/toxcore/dht/dht_friend.rs
```diff
@@ -57,54 +58,41 @@ impl DhtFriend {
...
+                }
+            }
+
+            if let Some(v4) = node.assoc4.ret_saddr {
+                if !node.assoc4.is_bad() {
+                    addrs.push(SocketAddr::V4(v4));
+                }
+            }
+
+            if self.pk == node.pk && !node.is_bad() {
```
kpp: `node` cannot be bad since we have already checked it in `.filter`

src/toxcore/dht/dht_node.rs
```diff
@@ -110,17 +155,35 @@ impl DhtNode {
...
-                    }
+                    self.assoc4.saddr.map(Into::into)
                 },
             }
         } else {
-            match self.saddr_v4 {
-                Some(v4) => Some(SocketAddr::V4(v4)),
-                None => None,
-            }
+            self.assoc4.saddr.map(Into::into)
```
kpp: if asso4 is empty we should try to check whether assoc6 is IPv6-mapped addr

NamsooCho: https://github.com/tox-rs/tox/blob/966afc3f921b6f9a765a6b819402ca9fcaaddc83/src/toxcore/dht/dht_node.rs#L88-L97

DhtNode always has at least one socket address.
So if IPv6 is None, IPv4 must be Some. Vice versa.
And If IPv6 is ipv4 mapped address, it is converted to IPv4 and stored in DhtNode.

kpp: OK

PR #185, opened 2018-08-03
feat(license): relicense to MIT/GPLv3+

kpp:
Fix https://github.com/tox-rs/tox/pull/162


Cargo.toml
```diff
@@ -17,7 +17,7 @@ documentation = "https://docs.rs/tox/"
 repository = "https://github.com/tox-rs/tox/"
 keywords = ["p2p", "instant-messaging", "tox", "toxcore", "networking"]
 categories = ["multimedia"]
-license = "MIT"
+license = "MIT OR GPL-3.0+"
```
kpp: @zetok do we have to use `AND` or `OR` condition here?

kpp:
From Yoni Rabkin via RT <licensing@fsf.org>:

```
Hello and thank you for writing in.

> I have some questions about gnu gpl v3+:

> 1. Do I have to put the license into COPYING file or can I name it
> LICENSE-GPL?

You can name the license file anything you wish.

> 2. Do I have to put GPLv3+ header into every source file of the
> project?  Can I avoid it?

It is recommended to add a copyright notice and license header to each
file, but it isn't a requirement. If the license information is in each
file, then someone who copies even a single file from the project still
has all of the relevant information about what they can, and cannot do
with the code.

These recommendations are like good coding conventions: they make it so
that someone looking at the project will immediately be familiar with
the licensing information and will have an easy time working with it.

I hope this is of help.
-- 
I am not a lawyer, the above is not legal advice
```

PR #186, opened 2018-08-03
Improve DHT tests

kurnevsky:


src/toxcore/dht/kbucket.rs
```diff
@@ -262,8 +262,8 @@ impl Bucket {
     }
 
     /// Check if node with given PK is in the `Bucket`.
-    pub fn contains(&self, pk: &PublicKey) -> bool {
-        self.nodes.iter().any(|n| &n.pk == pk)
+    pub fn contains(&self, base_pk: &PublicKey, pk: &PublicKey) -> bool {
+        self.nodes.binary_search_by(|n| base_pk.distance(&n.pk, pk)).is_ok()
```
kpp: Why would you do that? If you want to reduce N to log N then ... nobody cares because N = 8

kurnevsky: Why not? We use binary search everywhere but not here, why? I like consistency. Besides bucket not necessary has size of 8. Th biggest bucket we have at the moment - 32.

src/toxcore/dht/server/mod.rs
```diff
@@ -2385,7 +2400,19 @@ mod tests {
...
-    fn refresh_onion_key_test() {
+    fn handle_lan_discovery_when_disabled() {
+        let (mut alice, _precomp, _bob_pk, _bob_sk, _rx, addr) = create_node();
+
+        alice.enable_lan_discovery(false);
+        assert_eq!(alice.lan_discovery_enabled, false);
+
+        let lan = DhtPacket::LanDiscovery(LanDiscovery { pk: alice.pk });
+
+        assert!(alice.handle_packet(lan, addr).wait().is_ok());
```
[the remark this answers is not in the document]
kurnevsky: What do you mean by checking the result? It has type () so it can't be anything but empty.

src/toxcore/dht/kbucket.rs
kurnevsky: For instance this method uses binary search and takes base_pk: https://github.com/tox-rs/tox/blob/776a07f6e4d466fec562b1f1c879452baf4684b7/src/toxcore/dht/kbucket.rs#L156-L158
So when I stumbled when tried to use `contains` method with 2 keys in tests - suddenly it took only one key. So for consistency I decided to change it.

kpp: OK, let it be

src/toxcore/dht/server/mod.rs
```diff
@@ -2356,36 +2406,49 @@ mod tests {
...
-    fn refresh_onion_key_test() {
+    fn handle_lan_discovery_when_disabled() {
+        let (mut alice, _precomp, _bob_pk, _bob_sk, _rx, addr) = create_node();
+
+        alice.enable_lan_discovery(false);
+        assert_eq!(alice.lan_discovery_enabled, false);
+
+        let lan = DhtPacket::LanDiscovery(LanDiscovery { pk: alice.pk });
+
+        alice.handle_packet(lan, addr).wait().unwrap();
```
kpp: As I said before, you should also check that no actions are done when lan is disabled

kpp: rx is empty

PR #188, opened 2018-08-09
fix(clippy): fix clippy warnings

kpp:


src/toxcore/tcp/server/server_ext.rs
```diff
@@ -61,9 +61,9 @@ impl ServerExt for Server {
                     trace!("Send TCP packet {:?} to {:?}", client_pk, packet);
```
kurnevsky: Could you fix ordering here as well?

kpp: I will in another PR

kurnevsky: Don't forget then :)

PR #192, opened 2018-08-14
Change bootstrap method to save bootstrap nodes then retry sending

NamsooCho:


src/toxcore/dht/server/mod.rs
```diff
@@ -53,7 +53,8 @@ pub const PING_TIMEOUT: u64 = 5;
 pub const MAX_TO_PING: u8 = 32;
 /// How often in seconds to ping newly announced nodes.
 pub const TIME_TO_PING: u64 = 2;
-
```
kurnevsky: I'd left newline here.

src/toxcore/dht/server/mod.rs
```diff
@@ -246,13 +252,62 @@ impl Server {
...
+        let mut bootstrap_futures = self.initial_bootstrap.iter().cloned()
+            .map(|node| self.send_nodes_req(&mut request_queue, &node.into(), self.pk))
+            .collect::<Vec<_>>();
+
+        let close_nodes = self.close_nodes.read();
+        let mut futures = close_nodes.iter()
+            .map(|node| self.send_nodes_req(&mut request_queue, &node, self.pk))
+            .collect::<Vec<_>>();
+
+        futures.append(&mut bootstrap_futures);
```
kpp: I would say you should use `::extend`

kpp:
> Overall coverage decreased (-0.2%) to 96.149%

How about fixing the coverage?

src/toxcore/dht/server/mod.rs

src/toxcore/dht/server/mod.rs
```diff
@@ -246,13 +252,62 @@ impl Server {
...
+    /// until at least one entry inserted to Kbucket
+    /// and send NodesRequest to nodes in Kbucket if all nodes are discarded.
+    /// All nodes in Kbucket are discarded means that network is not available for a while.
+    fn run_recover_network_error(self) -> IoFuture<()> {
+        let interval = Duration::from_secs(BOOTSTRAP_INTERVAL);
+        let wakeups = Interval::new(Instant::now(), interval);
+        let self_c = self.clone();
+        let future = wakeups
+            .map_err(|e| Error::new(ErrorKind::Other, format!("Bootstrap timer error: {:?}", e)))
+            .take_while(move |_| future::ok(self.is_kbucket_empty() || self.is_all_discarded()))
```
kpp: > This function, like Iterator::take_while, will take elements from the stream until the predicate resolves to false. Once one element returns false it will always return that the stream is done.

Oops.

PR #193, opened 2018-08-14
NAT fixes

kurnevsky:


kpp:
Reduce the number of commits please

PR #200, opened 2018-08-23
Treat IPv4 and IPv6 addresses of same node as they belonged to different nodes

kurnevsky:


src/toxcore/dht/dht_node.rs
```diff
@@ -64,6 +64,20 @@ impl<T> SockAndTime<T> {
     pub fn is_ping_interval_passed(&self) -> bool {
         self.last_ping_req_time.map_or(true, |time| clock_elapsed(time) >= Duration::from_secs(PING_INTERVAL))
     }
+
+    /// Get address if it should be pinged and update `last_ping_req_time`.
+    pub fn ping_addr(&mut self) -> Option<T> {
+        if let Some(saddr) = self.saddr {
```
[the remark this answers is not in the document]
kurnevsky: Nope, this will return `Some` in all cases when `self.saddr` is some. I need `None` if the condition `!self.is_discarded() && self.is_ping_interval_passed()` is not satisfied.

src/toxcore/dht/dht_node.rs
```diff
@@ -37,7 +37,7 @@ pub struct SockAndTime<T> {
     pub ret_last_resp_time: Option<Instant>,
 }
 
-impl<T> SockAndTime<T> {
+impl<T: Copy> SockAndTime<T> {
```
kpp: How about `+ ToSocketAddrs`?

src/toxcore/dht/server/mod.rs
```diff
@@ -289,29 +290,22 @@ impl Server {
...
-            .map(|node| self.send_nodes_req(&mut request_queue, &node.into(), self.pk))
-            .collect::<Vec<_>>();
-
         let close_nodes = self.close_nodes.read();
-        let mut futures = close_nodes.iter()
-            .map(|node| self.send_nodes_req(&mut request_queue, &node, self.pk))
+        let futures = self.initial_bootstrap
+            .iter()
+            .cloned()
+            .chain(close_nodes
```
kpp: better change the order of close_nodes with initial_bootstrap to make the code a little bit prettier

src/toxcore/dht/dht_node.rs
kurnevsky: Done.

src/toxcore/dht/server/mod.rs
kurnevsky: Done.

src/toxcore/dht/server/mod.rs
```diff
@@ -430,10 +428,19 @@ impl Server {
         where T: Iterator<Item = &'a mut DhtNode>
     {
         let futures = nodes
-            .filter(|node| !node.is_discarded() && node.is_ping_interval_passed())
+            .flat_map(|node| {
+                node.assoc4
+                    .ping_addr()
```
kpp: Add temp variables for node_v4 and node_v6 and chain them

src/toxcore/dht/daemon_state.rs
```diff
@@ -59,9 +59,10 @@ impl DaemonState {
 
         let mut request_queue = server.request_queue.write();
         let nodes_sender = nodes.iter()
-            .map(|node|
-                server.send_nodes_req(&mut request_queue, &(*node).into(), server.pk)
-            );
+            .map(|node| {
+                let ping_id = request_queue.new_ping_id(node.pk);
```
NamsooCho: This line is redundant. Is there any reason for avoiding to pass request_queue as an argument?

kurnevsky: I made it to be consistent with `send_ping_req` but you are right - changed both to taking `request_queue`.

PR #202, opened 2018-08-27
Renaming and minor refactoring

kurnevsky:


src/toxcore/dht/server/mod.rs
```diff
@@ -266,35 +274,35 @@ impl Server {
...
-    /// Run initial bootstapping, It continues to sending NodesRequest to bootstrap nodes
-    /// until at least one entry inserted to Kbucket
-    /// and send NodesRequest to nodes in Kbucket if all nodes are discarded.
-    /// All nodes in Kbucket are discarded means that network is not available for a while.
-    fn run_recover_network_error(self) -> IoFuture<()> {
+    /// Run initial bootstrapping. It sends `NodesRequest` packet to bootstrap
+    /// nodes periodically if all nodes in Kbucket are discarded (including the
+    /// case when it's empty). It has to be an endless loop because we might
+    /// loose the network connection and thereby loose all nodes in Kbucket.
+    fn run_bootstrap_requests_sending(self) -> IoFuture<()> {
```
kpp: oh wait, why do we accept self by value?

src/toxcore/dht/server/mod.rs
```diff
@@ -266,35 +274,35 @@ impl Server {
...
-    /// Send NodesRequest to initial bootstapping nodes and nodes in Kbucket.
-    fn send_bootstrap_and_ping(&self) -> IoFuture<()> {
+    /// Check if all nodes in Kbucket are discarded (including the case when
+    /// it's empty) and if so then send `NodesRequest` packet to nodes from
+    /// initial bootstrap list and from Kbucket.
+    fn send_bootstrap_requests(&self) -> IoFuture<()> {
         let mut request_queue = self.request_queue.write();
         let close_nodes = self.close_nodes.read();
 
+        if !close_nodes.is_all_discarded() {
```
kpp:
```
fn main() {
    let x = [];
    let c = x.iter().all(|x: &i32| *x == 0);
    println!("c = {}", c);
}
ёёё

Comment (kpp) [reply src/toxcore/dht/server/mod.rs]: are you sure we should not check `close_nodes.is_empty()`?

Comment (kurnevsky) [reply src/toxcore/dht/server/mod.rs]: Yes, we should send request if `close_nodes.is_empty` OR `close_nodes.is_all_discarded`. When close nodes is_empty the second condition will be always `true`.

File: src/toxcore/dht/server/mod.rs [hunk 3]
```diff
@@ -3082,21 +3090,42 @@ mod tests {
...
+        alice.try_add_to_close_nodes(&pn);
+
+        // test with ipv6 mode
+        alice.enable_ipv6_mode(true);
+        alice.send_bootstrap_requests().wait().unwrap();
+
+        // Necessary to drop tx so that rx.collect() can be finished
+        drop(alice);
+
+        assert!(rx.collect().wait().unwrap().is_empty());
```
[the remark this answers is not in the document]
kurnevsky: We shouldn't send anything if kbucket has good nodes. I moved the condition inside this function so rx should be empty in this case.

src/toxcore/dht/server/mod.rs
kurnevsky: Because it's consumed inside. I believe if something is consumed it's supposed to be passed by value rather then by reference with copying inside.

kpp: > Because it's consumed inside

why?

kurnevsky: Because it has to be moved inside endless loop: https://github.com/tox-rs/tox/blob/818c22a56442181276e62d887f822205f7ec9f16/src/toxcore/dht/server/mod.rs#L284
It can't be used by reference there. So either we take it by value or clone inside.

kpp: Thanks

PR #205, opened 2018-08-30
Parser of Share relays

NamsooCho:


src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ShareRelays {
+    relays: Vec<TcpUdpPackedNode>,
+}
+
+impl FromBytes for ShareRelays {
+    named!(from_bytes<ShareRelays>, do_parse!(
+        tag!("\x11") >>
+        relays: many1!(TcpUdpPackedNode::from_bytes) >>
+        verify!(value!(relays.len()), |len| len == 3) >>
```
kurnevsky: See https://docs.rs/nom/3.2.1/nom/macro.count.html

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,82 @@
...
+            relays,
+        })
+    ));
+}
+
+impl ToBytes for ShareRelays {
+    fn to_bytes<'a>(&self, buf: (&'a mut [u8], usize)) -> Result<(&'a mut [u8], usize), GenError> {
+        do_gen!(buf,
+            gen_be_u8!(0x11) >>
+            gen_many_ref!(&self.relays, |buf, relay| TcpUdpPackedNode::to_bytes(relay, buf))
```
kurnevsky: Maybe check relays count? Like it was done here: https://github.com/tox-rs/tox/blob/e9c8174dbc804e167810589ab33f584c3ff448c5/src/toxcore/dht/packet/nodes_response.rs#L140

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,82 @@
+/*! ShareRalays struct
+*/
+
+use toxcore::binary_io::*;
+use toxcore::state_format::old::*;
+
+/** ShareRelays is a struct that holds 3 TCP relays as a PackedNode format.
+
+This packet is used to share relay info. between two friends.
```
kurnevsky: Unnecessary dot in the middle.

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,82 @@
...
+
+Length    | Content
+--------- | ------
+`1`       | `0x11`
+variable  | 3 PackedNodes
+
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ShareRelays {
+    relays: Vec<TcpUdpPackedNode>,
```
kurnevsky: Why not just `PacketNode`?

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,82 @@
...
+/** ShareRelays is a struct that holds 3 TCP relays as a PackedNode format.
+
+This packet is used to share relay info. between two friends.
+
+Serialized form:
+
+Length    | Content
+--------- | ------
+`1`       | `0x11`
+variable  | 3 PackedNodes
```
kurnevsky: Length is not variable in this case.

src/toxcore/friend_connection/packet/share_relays.rs
kurnevsky: Is it possible that this packet contains UDP addresses? If not, then can we create an alternative PackedNode but for TCP addresses only?

kurnevsky: Or maybe just add different methods for serialize/deserialize PacketNode with TCP and with UDP address?

src/toxcore/friend_connection/packet/share_relays.rs
NamsooCho: Size of PackedNode is variable, so size of this struct is variable.

src/toxcore/friend_connection/packet/share_relays.rs
kurnevsky: ok, got it.

src/toxcore/friend_connection/packet/share_relays.rs
kurnevsky: ok, got it.

src/toxcore/friend_connection/packet/share_relays.rs
kurnevsky: Size of serialized array has nothing to do with our representation of packet struct. We can still have fixed size array since we know the count. It's not always convenient due to limitations of such arrays but maybe here we can use them. Or if not - then just add check for the _count_ (not length).

src/toxcore/friend_connection/packet/share_relays.rs
NamsooCho: There are two options.
PackedNode of UDP, TCP, TCP_UDP, 3 separated struct is one option.
One struct which has different methods for TCP and UDP and TcpUdp.
One struct way seems better, but this case, other part of source code should be changed.

kurnevsky: But is seems you need only TCP addresses here, right? Then it seems to me that it's better to reuse PackedNode and change/add parsing/serializing methods.

NamsooCho: Ok.
I will try to do that you said.

src/toxcore/friend_connection/packet/share_relays.rs
NamsooCho: The array size should be fixed when compile time. Right?
I think that array is not dynamic memory method.
The size of this struct is variable when run-time.

kurnevsky: But you wrote that there are always 3 nodes in this struct: https://github.com/tox-rs/tox/pull/205/files#diff-b046e67b27fca0a355ab477d4a703491R16

So why fixed array with a size of 3 can't be used?

src/toxcore/friend_connection/packet/share_relays.rs
kurnevsky: I looked through our state format implementation - it seems we don't need `TcpUdpPackedNode` there as well - we know at compile time what type should be used (no need to fix it now if anything).

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ShareRelays {
+    relays: Vec<PackedNode>,
+}
+
+impl FromBytes for ShareRelays {
+    named!(from_bytes<ShareRelays>, do_parse!(
+        tag!("\x11") >>
+        relays: many1!(PackedNode::from_tcp_bytes) >>
+        verify!(value!(relays.len()), |len| len == 3) >>
```
kurnevsky: Use `count`.

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,68 @@
...
+            relays: relays,
+        })
+    ));
+}
+
+impl ToBytes for ShareRelays {
+    fn to_bytes<'a>(&self, buf: (&'a mut [u8], usize)) -> Result<(&'a mut [u8], usize), GenError> {
+        do_gen!(buf,
+            gen_be_u8!(0x11) >>
+            gen_cond!(self.relays.len() != 3, |buf| gen_error(buf, 0)) >>
```
kpp: Does it mean that we have to share exactly 3 relays?

NamsooCho: It should have at most 3 relays.

src/toxcore/friend_connection/packet/share_relays.rs
```diff
@@ -0,0 +1,69 @@
...
+    relays: Vec<PackedNode>,
+}
+
+impl FromBytes for ShareRelays {
+    named!(from_bytes<ShareRelays>, do_parse!(
+        tag!("\x11") >>
+        relays: many0!(PackedNode::from_tcp_bytes) >>
+        verify!(value!(relays.len()), |len| len <= 3) >>
+        (ShareRelays {
+            relays: relays,
```
kurnevsky: `relays: relays,` -> `relays,`

NamsooCho: done

src/toxcore/friend_connection/packet/share_relays.rs
kpp: Then the condition should be `<= 3`, shouldn't it?

NamsooCho: > gen_cond!(bool, I -> Result<I,E>) => I -> Result<I,E> Conditional combinator

if condition is true then closure runs. So if len() > 3 then generate error.

PR #207, opened 2018-09-02
Simplify constructor of Kbucket

kurnevsky:


src/toxcore/dht/kbucket.rs
```diff
@@ -103,38 +97,18 @@ pub struct Bucket {
...
-                trace!("Creating a new Bucket with capacity: {}", n);
-                Bucket {
-                    capacity: n,
-                    nodes: Vec::with_capacity(n as usize),
-                }
-            }
+    pub fn new(capacity: u8) -> Self {
+        trace!("Creating a new Bucket with capacity: {}", capacity);
+        Bucket {
+            capacity,
```
NamsooCho: Do we need this field `capacity`?

NamsooCho: Regarding clone(), I agree on keeping this member variable. Bucket and Kbucket could omit deriving Clone, but it may be a problem in the future.

src/toxcore/dht/server/mod.rs
```diff
@@ -187,10 +189,10 @@ impl Server {
             onion_symmetric_key: Arc::new(RwLock::new(secretbox::gen_key())),
             onion_announce: Arc::new(RwLock::new(OnionAnnounce::new(pk))),
             friends: Arc::new(RwLock::new(friends)),
-            nodes_to_bootstrap: Arc::new(RwLock::new(Bucket::new(None))),
+            nodes_to_bootstrap: Arc::new(RwLock::new(Bucket::new(MAX_TO_BOOTSTRAP))),
```
kpp: Why do we keep them in a Bucket instead of HashSet?

NamsooCho: We use try_add() of Bucket.

kpp: Then we don’t have to use try_bucket. What's the point?

kurnevsky: Because we can't afford to store all nodes - it can lead to memory overflow. So we have to limit the size somehow. c-toxcore leaves only the closest nodes, so do we. But I'm not sure if it's correct and how it's related to the eclipse attack. Anyway it's the topic for another PR - I don't like that we use for these lists full DhtNode either so I'll try to change it somehow (though not sure yet how).

PR #209, opened 2018-09-04
Speed up 'run' test

kurnevsky:


kpp:
How does it works?

kurnevsky:
In this test we send `PingRequest` from a client to a server, wait for a response, and then wait for a `PingRequest` from the server to the client. Because of that last step we wait 30 seconds in this test. I just used mocked time and set time right after receiving a response on the previous step so the server gets this time and sees that the client should be pinged.

PR #210, opened 2018-09-04
Add parameter for eviction strategy

kurnevsky:


src/toxcore/dht/dht_node.rs
```diff
@@ -234,17 +234,17 @@ mod tests {
...
             let mut bucket = Bucket::new(BUCKET_DEFAULT_SIZE);
-            assert_eq!(true, bucket.try_add(&pk, &n1));
-            assert_eq!(true, bucket.try_add(&pk, &n2));
-            assert_eq!(true, bucket.try_add(&pk, &n3));
-            assert_eq!(true, bucket.try_add(&pk, &n4));
-            assert_eq!(true, bucket.try_add(&pk, &n5));
-            assert_eq!(true, bucket.try_add(&pk, &n6));
-            assert_eq!(true, bucket.try_add(&pk, &n7));
-            assert_eq!(true, bucket.try_add(&pk, &n8));
+            assert_eq!(true, bucket.try_add(&pk, &n1, true));
```
kpp: what does `true` mean?

kurnevsky: This test should be removed - it tests nothing in this module.

src/toxcore/dht/kbucket.rs
```diff
@@ -477,15 +501,20 @@ impl Kbucket {
     }
 
     /// Create iterator over [`DhtNode`](./struct.DhtNode.html)s in `Kbucket`.
+    /// Nodes that this iterator produces are sorted by distance to a base
+    /// `PublicKey` (in ascending order).
     pub fn iter(&self) -> impl Iterator<Item = &DhtNode> {
         self.buckets.iter()
+            .rev()
```
kpp: Why rev?

kurnevsky: Because kbucket stores farther nodes at the beginning, but bucket - at the end.

PR #212, opened 2018-09-06
feat(dht): fill node_to_bootstrap of friend with close nodes

NamsooCho:


src/toxcore/dht/server/mod.rs
```diff
@@ -220,9 +220,16 @@ impl Server {
     }
 
     /// Add a friend.
-    pub fn add_friend(&self, friend: DhtFriend) {
+    /// `node_to_bootstrap` of new friend is filled with close nodes for fast bootstrapping.
+    pub fn add_friend(&self, mut friend: DhtFriend) {
         let mut friends = self.friends.write();
 
+        let close_nodes = self.close_nodes.read().get_closest(&friend.pk, true);
```
kurnevsky: I suggest to unify with: https://github.com/tox-rs/tox/blob/64d49ce2730ff30826ef9f33b66824f24c3bda88/src/toxcore/dht/server/mod.rs#L770-L792

i.e. create method in the server that gets closest nodes from both `close_nodes` and `friends` and use it here as well.

src/toxcore/dht/server/mod.rs
```diff
@@ -220,9 +220,16 @@ impl Server {
     }
 
     /// Add a friend.
-    pub fn add_friend(&self, friend: DhtFriend) {
+    /// `node_to_bootstrap` of new friend is filled with close nodes for fast bootstrapping.
+    pub fn add_friend(&self, mut friend: DhtFriend) {
```
kurnevsky: I suggest to create `DhtFriend` inside this method and take `PublicKey` only. We don't need to do anything with a `DhtFriend` outside and it can be hidden.

src/toxcore/dht/server/mod.rs
```diff
@@ -2664,13 +2675,15 @@ mod tests {
...
             PackedNode::new("127.1.1.4:12345".parse().unwrap(), &gen_keypair().0),
         ];
-        let mut friend = DhtFriend::new(friend_pk);
-        for node in &nodes {
-            friend.close_nodes.try_add(&friend_pk, &node);
-            let dht_node = friend.close_nodes.get_node_mut(&friend_pk, &node.pk).unwrap();
-            dht_node.update_returned_addr(node.saddr);
+        alice.add_friend(friend_pk);
+        {
+            let friend = &mut alice.friends.write();
```
kurnevsky: friend_s_

PR #214, opened 2018-09-10
feat(messenger): parser of Online packet

NamsooCho:


src/toxcore/friend_connection/packet/alive.rs
```diff
@@ -1,8 +1,8 @@
+
 /*! Alive struct
 */
 
 use toxcore::binary_io::*;
-
```
kurnevsky: No need to move the empty line to the beginning.

NamsooCho: done

PR #216, opened 2018-09-10
Simplify dht node

kurnevsky:
Changes:
- Since `Bucket` stores only nodes that we can reach `DhtNode` can have IPv6 address only if `is_ipv6_enabled` is `true`. It means that we don't have to pass this variable to `DhtNode` and store it in `Kbucket` anymore. So it was removed.

### DIFF
--- a/src/toxcore/dht/dht_node.rs
+++ b/src/toxcore/dht/dht_node.rs
@@ -128,62 +128,45 @@ impl DhtNode {
         self.assoc4.is_discarded() && self.assoc6.is_discarded()
     }
 
-    /// Return SocketAddr for DhtNode
-    pub fn get_socket_addr(&self, is_ipv6_enabled: bool) -> Option<SocketAddr> {
-        let addr = if is_ipv6_enabled {
-            match self.assoc6.saddr {
-                Some(v6) => {
-                    match self.assoc4.saddr {
-                        Some(v4) => {
-                            if self.assoc4.last_resp_time >= self.assoc6.last_resp_time {
-                                Some(SocketAddr::V4(v4))
-                            } else {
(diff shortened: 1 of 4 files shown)


src/toxcore/dht/dht_node.rs
```diff
@@ -129,61 +129,43 @@ impl DhtNode {
...
         if addrs.is_empty() {
-            warn!("Can't find target address in DhtNode");
+            warn!("get_all_addrs: DhtNode doesn't have IP addresses!");
         }
 
         addrs
     }
 
     /// Convert `DhtNode` to `PackedNode` based on `is_ipv6_enabled` flag.
-    pub fn to_packed_node(&self, is_ipv6_enabled: bool) -> Option<PackedNode> {
```
kpp: Fix doc comment

src/toxcore/dht/server/mod.rs
```diff
@@ -207,7 +207,6 @@ impl Server {
     /// Enable/disable IPv6 mode of DHT server.
     pub fn enable_ipv6_mode(&mut self, enable: bool) {
         self.is_ipv6_enabled = enable;
```
kpp: Do we still need it?

kurnevsky: It's used to filter out received nodes with IPv6 address if we don't support it: https://github.com/tox-rs/tox/blob/f5826d7f88bde1860100895a16cffa424297cc33/src/toxcore/dht/server/mod.rs#L834-L836
Otherwise we'd add it to the `nodes_to_bootstrap` list and then try to ping without success. So it'd only take up a place in `nodes_to_bootstrap` with probably other good nodes eviction.

PR #220, opened 2018-09-11
Validate PublicKey when parsing

kurnevsky:


kpp:
Did you check that we use PublicKey::from_bytes every time we need to parse it?

NamsooCho:
It would be better if we can do valid-checking one time when PK get in our library.

kurnevsky:
Try grep by 'PublicKey::from_slice' - there are no usages besides tests.

PR #221, opened 2018-09-12
Give appropriate names to the bucket and kbucket

kurnevsky:
Renamings:
Kbucket -> Ktree (refers to kademlia binary tree)
Bucket -> Kbucket


kpp:
Please add link to kademlia paper in pdf into doc comments.

PR #227, opened 2018-09-17
Use tokio::spawn for running TCP connections

kurnevsky:


src/toxcore/tcp/server/server_ext.rs
```diff
@@ -31,24 +32,32 @@ impl ServerExt for Server {
     fn run(self: Self, listner: TcpListener, dht_sk: SecretKey) -> IoFuture<()> {
```
kpp: Add notice that ServerExt::run should run inside `tokio::run`

kurnevsky: done

PR #230, opened 2018-09-18
Get rid of quickcheck

kurnevsky:


src/toxcore/state_format/old.rs
```diff
@@ -43,7 +43,7 @@ pub const NOSPAMKEYSBYTES: usize = NOSPAMBYTES + PUBLICKEYBYTES + SECRETKEYBYTES
 /// The `Default` implementation generates random `NospamKeys`.
 impl Default for NospamKeys {
     fn default() -> Self {
-        let nospam = NoSpam::default();
+        let nospam = NoSpam::new();
```
kpp: Actually `new` should be implemented from `default`

kurnevsky: There is no default value for `NoSpam`. `NoSpam::new` generates a random `NoSpam`. `Default` trait is supposed to give the same 'default' value every call, not random.

kpp: Did you get warnings from clippy about that?

kurnevsky: We have these warnings disabled:
```rust
#![cfg_attr(feature = "cargo-clippy", allow(clippy::new_without_default, clippy::new_without_default_derive))]
```
Actually I'm unsure regarding these default implementations. But I'd rather rename `new` to `random` and not provide `Default` trait implementation. Then we can reenable these warnings.

kpp: OK. In another PR.

PR #231, opened 2018-09-19
feat(tcp_conn): add data structure for tcp_connection and ctor

NamsooCho:


src/toxcore/tcp/connections.rs
```diff
@@ -0,0 +1,213 @@
...
+    Sleeping,
+}
+
+/// Main struct for TCP connections.
+/// Holds key_pair for TCP connections, set of TCP connections to relay,
+/// set of connections of TcpClient.
+/// PublicKey is used as a key of HashMap to avoid ABA problem.
+pub struct Connections<'a> {
+    /// Dht Server.
+    dht: &'a Server,
```
kurnevsky: Are you sure that you need a reference? Where are you going to store these connections? DHT server is cloneable so it might be easier to store just a `Server` object.

src/toxcore/tcp/connections.rs
```diff
@@ -0,0 +1,213 @@
...
+    sk: SecretKey,
+    conns_of_client: Arc<RwLock<HashMap<PublicKey, ConnOfClient<'a>>>>,
+    connections: Arc<RwLock<HashMap<PublicKey, Connection>>>,
+}
+
+/// Connection of a TcpClient object.
+pub struct ConnOfClient<'a> {
+    status: ConnOfClientStatus,
+    /// Tcp Client.
+    client: &'a Client,
```
kurnevsky: Where the `Client` is going to be stored? And the `ConnOfClient` so it can store the reference with specified lifetime?

src/toxcore/tcp/connections.rs
kpp: What's a Dht server for?

NamsooCho: Now, can't find the user of dht in c-toxcore. So it is removed in struct

src/toxcore/tcp/connections.rs
NamsooCho: TcpClient is created when add tcp_relay_connection is called in NetCrypto.
Also, in TCP_connections it is created when reconnect to a relay or during wakeup.

kurnevsky: But in order to store the reference in `ConnOfClient` you need to store the `Client` somewhere, right? So where would it be?

src/toxcore/tcp/connections.rs
```diff
@@ -0,0 +1,195 @@
...
+When a Tox client connects to a friend via tcp relay, normally 3 redundant connections are established.
+One connection is used for data send/recv, 2 others are backup.
+In toxcore maximum number of redundant connections is 6.
+TCP connection can get into sleep mode.
+Getting into sleep mode can occur when UDP connecion is established, because Tox prefer UDP than TCP.
+When established UDP connection is disabled, TCP connecions are awaken.
+
+Defintion of struct names.
+
+<pre style="white-space:pre;">
```
kurnevsky: Maybe use usual markdown list?

src/toxcore/tcp/connections.rs
NamsooCho: done

src/toxcore/tcp/connections/connection.rs
```diff
@@ -0,0 +1,411 @@
+/*! The implementation of TCP connections
```
kurnevsky: Why copy this file from the `client` module?

src/toxcore/tcp/connections/connections.rs
```diff
@@ -0,0 +1,319 @@
...
+    None,
+    /// Connection is registered internally.
+    Registered,
+    /// Connection is online.
+    Online,
+}
+
+/// Status of connection to a friend.
+#[derive(PartialEq, Clone)]
+pub enum ConnOfClientStatus {
```
kurnevsky: Since there is `ConnectionStatus` above it'd look prettier if this struct had a name like `ClientConnectionStatus`.

src/toxcore/tcp/connections/connections.rs
```diff
@@ -0,0 +1,319 @@
...
+pub struct Connections {
+    pk: PublicKey,
+    sk: SecretKey,
+    conns_of_client: Arc<RwLock<HashMap<PublicKey, ConnOfClient>>>,
+    connections: Arc<RwLock<HashMap<PublicKey, Connection>>>,
+}
+
+/// Connection of a TcpClient processor object.
+#[derive(Clone)]
+pub struct ConnOfClient {
```
kurnevsky: ClientConnection?

src/toxcore/tcp/connections/connections.rs
```diff
@@ -0,0 +1,319 @@
...
+        let reader = from_server
+            .map_err(Error::from)
+            .map(move |packet| (packet, relay.clone().id_of_client))
+            .forward(from_server_tx
+                .sink_map_err(|e| {
+                    err_msg(format!("Could not forward message from server to connection {:?}", e))
+                })
+            )
+            .map(|_| {
+                println!("Connection closed");
```
kurnevsky: No need to print it.

src/toxcore/tcp/connections/connection.rs
NamsooCho: https://github.com/tox-rs/tox/blob/f914486e203191526dc9dfb6f10d03fa31ff5498/src/toxcore/tcp/connections/connection.rs#L59

PublicKey is added.

NamsooCho: It is used as a connection_id

src/toxcore/tcp/connections/connections.rs
NamsooCho: done

PR #238, opened 2018-09-25
Use callback to get the message of the day

kurnevsky:


src/toxcore/dht/server/mod.rs
```diff
@@ -195,7 +196,7 @@ impl Server {
             last_nodes_req_time: Arc::new(RwLock::new(clock_now())),
             nodes_to_ping: Arc::new(RwLock::new(NodesQueue::new(MAX_TO_PING))),
             tox_core_version: 0,
-            motd: Vec::new(),
+            motd_cb: Arc::new(|_| Vec::new()),
```
kpp: How about using Option? Clients do not have a handler for BootstrapInfo

kurnevsky: You mean if it's None then don't respond to BootstrapInfo packets?

kpp: Yes

PR #239, opened 2018-10-03
Use failure for Codec and ServerExt

kurnevsky:


src/toxcore/tcp/server/server_ext.rs
```diff
@@ -100,19 +117,19 @@ impl ServerExt for Server {
                 .for_each(move |packet| {
                     debug!("Handle {:?} => {:?}", client_pk, packet);
                     server_c_c.handle_packet(&client_pk, packet)
+                        .map_err(|e| e.into())
                 });
 
             // writer = for each Packet from to_client_rx send it to client
             let writer = to_client_rx
-                .map_err(|()| Error::from(ErrorKind::UnexpectedEof))
+                .map_err(|()| ConnectionError::ClientStreamError.into())
```
kurnevsky: Maybe use `err_msg` here since this error should not be happening?

src/toxcore/tcp/server/server_ext.rs
```diff
@@ -66,21 +86,18 @@ impl ServerExt for Server {
...
+            }.into())),
         };
 
         debug!("A new TCP client connected from {}", addr);
 
         let register_client = make_server_handshake(stream, dht_sk.clone())
-            .map_err(|e|
-                Error::new(ErrorKind::Other, format!("Handshake error: {}", e))
-            )
+            .map_err(|e| e.into())
```
kurnevsky: Should we wrap such errors to `ConnectionError`? `make_server_handshake` will return it's own `failure::Error` so I decided to return it as is. What do you think?

src/toxcore/tcp/server/server_ext.rs
```diff
@@ -1,41 +1,62 @@
...
+        error: TimeoutError<Error>
+    },
+}
+
 /// Extension trait for running TCP server on incoming `TcpStream` and ping sender
 pub trait ServerExt {
     /// Running TCP ping sender and incoming `TcpStream`. This function uses
     /// `tokio::spawn` inside so it should be executed via tokio to be able to
     /// get tokio default executor.
-    fn run(self: Self, listner: TcpListener, dht_sk: SecretKey) -> IoFuture<()>;
```
kpp: Why did you replace `IoFuture` with `Box<...>`?

src/toxcore/tcp/server/server_ext.rs
kpp: Usually it is `.map_err(|()| unreachable!("rx can't fail"))` like in https://github.com/jgallagher/tokio-chat-example/blob/master/tokio-chat-server/src/main.rs#L198

src/toxcore/tcp/server/server_ext.rs
```diff
@@ -1,41 +1,62 @@
...
+    PeerAddrError {
+        /// Peer address error
+        error: IoError,
+    },
+    /// Error indicates that client stream was terminated. Should not happen
+    #[fail(display = "Client stream terminated")]
+    ClientStreamError,
+    /// Sending packet error. Can be either timeout or underlying sending error
+    #[fail(display = "Failed to send TCP packet: {}", error)]
+    SendPacketError {
```
kpp: How about renaming it to something with `timeout` in the name?

src/toxcore/tcp/server/server_ext.rs
kpp: Would you try it locally before merge? Will the code compile?

kurnevsky: It already compiles, but later when `make_server_handshake` returns `failure::Error` we will be able to remove this `map_err`.

src/toxcore/tcp/server/server_ext.rs
kurnevsky: Because now it's different type. Error is not `std::io::Error` but `failure::Error`. Later we could use existential types here.

src/toxcore/tcp/server/server_ext.rs
kurnevsky: It's not necessary timeout - it might be serialization error, or some other underlying `faulure::Error`.

kurnevsky: See https://docs.rs/tokio-timer/0.2.7/src/tokio_timer/timeout.rs.html#79
It can contain `Kind::Inner`.

src/toxcore/tcp/codec.rs
```diff
@@ -43,19 +81,18 @@ impl Decoder for Codec {
...
         // deserialize Packet
         match Packet::from_bytes(&decrypted_data) {
-            IResult::Incomplete(_) => {
-                Err(Error::new(ErrorKind::Other,
-                    "Packet should not be incomplete"))
+            IResult::Incomplete(needed) => {
+                Err(DecodeError::IncompleteDecryptedPacket {
+                    len: decrypted_data.len(),
+                    needed
+                }.into())
```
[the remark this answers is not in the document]
kurnevsky: `run_connection` returns `Box<Future>` so `?` can't be used here.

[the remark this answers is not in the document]
kpp: But what about Codec::decode?

kpp:
> Overall coverage decreased (-0.3%) to 96.169%

Why?

PR #241, opened 2018-10-05
Use failure for toxencryptsave

kurnevsky:


src/toxencryptsave/mod.rs
```diff
@@ -167,7 +165,7 @@ impl PassKey {
                               // ↓ don't
     let passkey = PassKey::new(&[0]).expect("Failed to unwrap PassKey!");
 
-    assert_eq!(Err(EncryptionError::Null), passkey.encrypt(&[]));
+    assert_eq!(passkey.encrypt(&[]), Err(EncryptionError::Null));
```
kpp: Why did you change the order of rhs & lhs?

kurnevsky: We mostly use this ordering. Also it used to be in rust src: https://github.com/rust-lang/rust/pull/19647/files
But can revert it - it's not too important.

kpp: Right!

PR #245, opened 2018-10-20
Add parser of hardening DhtRequest

NamsooCho:


src/toxcore/dht/packet/dht_request.rs
```diff
@@ -130,6 +131,10 @@ pub enum DhtRequestPayload {
     NatPingResponse(NatPingResponse),
     /// [`DhtPkAnnounce`](./struct.DhtPkAnnounce.html) structure.
     DhtPkAnnounce(DhtPkAnnounce),
+    /// [`HardenRequest`](./struct.HardenRequest.html) structure.
+    HardenRequest(HardenRequest),
```
kpp: Why `Harden` not `Hardening`?

NamsooCho: done

src/toxcore/dht/packet/dht_request.rs
```diff
@@ -264,9 +273,89 @@ impl ToBytes for DhtPkAnnounce {
...
+    pub node: PackedNode,
+    /// Search id
+    pub id: PublicKey,
+}
+
+impl FromBytes for HardeningRequest {
+    named!(from_bytes<HardeningRequest>, do_parse!(
+        tag!("\x30") >>
+        tag!("\x02") >>
+        node: call!(PackedNode::from_bytes) >>
```
kurnevsky: It's not PackedNode format, they just use memcpy of plain c struct:
https://github.com/TokTok/c-toxcore/blob/744dc2f5da2c54ad1e4d7d6ce229bc1756d81263/toxcore/DHT.c#L2442
So I'd suggest not to parse it (or parse as a byte array). Anyway it seems it's going to be removed from `c-toxcore`.

NamsooCho: https://github.com/TokTok/c-toxcore/blob/744dc2f5da2c54ad1e4d7d6ce229bc1756d81263/toxcore/DHT.c#L2439

It is actually parsed.

NamsooCho: https://github.com/TokTok/c-toxcore/blob/744dc2f5da2c54ad1e4d7d6ce229bc1756d81263/toxcore/DHT.c#L2439

It is actually parsed.

NamsooCho: https://github.com/TokTok/c-toxcore/blob/744dc2f5da2c54ad1e4d7d6ce229bc1756d81263/toxcore/DHT.c#L2439

It is actually parsed.

NamsooCho: https://github.com/TokTok/c-toxcore/blob/744dc2f5da2c54ad1e4d7d6ce229bc1756d81263/toxcore/DHT.c#L2439

It is actually parsed.

kpp: @NamsooCho unfortunately I removed all duplicate comments + the last one. Sorry)

Basically we don't care. All we need is id's and the length of the structs. We don't need to parse the underlying struct correctly, we need to create a packet for handling (fake fandling =) ))

kpp: @NamsooCho unfortunately I removed all duplicate comments + the last one. Sorry)

Basically we don't care. All we need is id's and the length of the structs. We don't need to parse the underlying struct correctly, we need to create a packet for handling (fake handling =) ))

kpp: Basically we don't care. All we need is id's and the length of the structs. We don't need to parse the underlying struct correctly, we need to create a packet for handling (fake handling =) ))

kpp: Basically we don't care. All we need is id's and the length of the structs. We don't need to parse the underlying struct correctly, we need to create a packet for handling (fake handling =) ))

kpp: Basically we don't care. All we need is id's and the length of the structs. We don't need to parse the underlying struct correctly, we need to create a packet for handling (fake handling =) ))

kpp: Basically we don't care. All we need is id's and the length of the structs. We don't need to parse the underlying struct correctly, we need to create a packet for handling (fake handling =) ))

NamsooCho: done

src/toxcore/dht/packet/dht_request.rs
```diff
@@ -264,9 +272,70 @@ impl ToBytes for DhtPkAnnounce {
...
+*/
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct HardeningResponse;
+
+impl FromBytes for HardeningResponse {
+    named!(from_bytes<HardeningResponse>, do_parse!(
+        tag!("\x30") >>
+        tag!("\x03") >>
+        rest >>
```
kurnevsky: It's not necessary to parse `rest` since we don't use it and don't require `eof`.

src/toxcore/dht/packet/dht_request.rs
```diff
@@ -264,9 +272,70 @@ impl ToBytes for DhtPkAnnounce {
...
+
+*/
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct HardeningRequest;
+
+impl FromBytes for HardeningRequest {
+    named!(from_bytes<HardeningRequest>, do_parse!(
+        tag!("\x30") >>
+        tag!("\x02") >>
+        rest >>
```
kurnevsky: It's not necessary to parse `rest` since we don't use it and don't require `eof`.

src/toxcore/dht/packet/dht_request.rs
kpp: That's a big question what shall we do here

kpp: I suggest to leave `rest` but we have to mark with comments what is going on. Also structs like HardeningResponse and HardeningRequest should be commented, that they are not used and they are depricated.

NamsooCho: done

PR #247, opened 2018-10-22
Move connection_id validation to parsers

kurnevsky:


kpp:
What if we will change binary protocol to JSON or any other? How would you check connection id?

kurnevsky:
> What if we will change binary protocol to JSON or any other? How would you check connection id?

Why would we do that? There is no reason to change protocol to json without changing everything else - relay should minimize delays for audio/video which means data overhead should be as little as possible. And even if we use json that means that everything else will be changed, so probably there will be no connection_id. Anyway it's something very far/improbable so I wouldn't consider it at the moment.

kpp:
Ok

PR #250, opened 2018-10-24
Add DHT packet handler to be handled in client only

NamsooCho:


src/toxcore/dht/server/hole_punching.rs
```diff
@@ -211,6 +211,12 @@ impl HolePunching {
 mod tests {
     use super::*;
 
+    #[test]
+    fn hole_punchiong_clonable() {
```
kpp: We don't mix features. So you have to create another PR for that test.

src/toxcore/dht/server/mod.rs
```diff
@@ -696,12 +696,30 @@ impl Server {
...
                     Error::new(ErrorKind::Other,
-                        format!("Packet is not handled {:?}", p)
-                )))
-            }
+                               format!("Packet is not handled {:?}", packet)
+                    )))
+            },
+            // This packet should be handled in client only
+            Packet::OnionDataResponse(packet) => {
+                debug!("This packet should be handled in client only, {:?}", packet.clone());
```
kurnevsky: No need to clone packet.

src/toxcore/dht/server/mod.rs
```diff
@@ -696,12 +696,30 @@ impl Server {
...
+                Box::new(future::err(
+                    Error::new(ErrorKind::Other,
+                               format!("Packet is not handled {:?}", packet)
+                    )))
+            },
+            // This packet should be handled in client only
+            Packet::OnionAnnounceResponse(packet) => {
+                debug!("This packet should be handled in client only, {:?}", packet.clone());
```
kurnevsky: No need to clone packet.

PR #253, opened 2018-10-26
Fix connection_id verification

kurnevsky:


kpp:
What about code coverage? I don’t see the report.

kurnevsky:
@kpp you can see it if you unfold the checks list: https://coveralls.io/builds/19748725

kpp:
`coverage decreased (-0.03%) to 95.144%` but that's fine.

PR #256, opened 2018-10-29
Fix NLL warnings

NamsooCho:


src/lib.rs
```diff
@@ -5,6 +5,9 @@ Repo: https://github.com/tox-rs/tox
 
 */
 
+// TODO: uncomment this feature when NLL is supported by stable rustc
+//#![feature(nll)]
```
kpp: Remove these comments. We will enable it soon, no need to keep TODO in the code

PR #257, opened 2018-10-30
DHT cache precomputed keys

NamsooCho:


src/toxcore/dht/server/mod.rs
```diff
@@ -208,9 +211,17 @@ impl Server {
             lan_discovery_enabled: true,
             is_ipv6_enabled: false,
             initial_bootstrap: Vec::new(),
+            precomputed_keys: Arc::new(RwLock::new(HashMap::new())),
         }
     }
 
+    /// Get PrecomputedKey of a peer or newly make and save it.
+    fn get_precomputed(&self, pk: PublicKey) -> PrecomputedKey {
+        let mut precomputed = self.precomputed_keys.write();
```
kpp: Well, better take `read` lock first, then `write` to insert missing precomputed key.

NamsooCho: The `entry()` is good for HashMap on this purpose.
One read lock and one write lock can not be used in same scope.

kpp: Then you should end the scope or `read` lock to get `write` lock.

src/toxcore/dht/server/mod.rs
```diff
@@ -74,6 +75,25 @@ struct ServerBootstrapInfo {
     motd_cb: Arc<Fn(&Server) -> Vec<u8> + Send + Sync>,
 }
 
+/// Struct that contains PrecomputedKey of a peer and last time it is accessed.
+#[derive(Clone)]
+pub struct PrecomputedKeyAndTime {
+    /// PrecomputedKey of a peer.
+    precomputed_key: PrecomputedKey,
+    /// last time of accessing precomputed_key.
+    pub last_time: Instant,
```
kpp: Why do you keep `last_time`? What's that for?

NamsooCho: It is for removing precomputed keys.
If last_time span is greater than KILL_NODE_TIMEOUT then it is removed.

src/toxcore/dht/server/mod.rs

src/toxcore/dht/server/mod.rs
kpp: You don't need to duplicate it. Once you remove the node you can remove the key by node' PK

NamsooCho: https://github.com/tox-rs/tox/blob/9309968c7ae68dccfc9b22a1687beea7989e67c2/src/toxcore/dht/server/mod.rs#L763-L765

When we handle ping_request we create and save precomputed key.
But it is possible we don't add peer into `nodes_to_ping` by `ping_add` if the peer can not be added to close_node.
This means that we just responds with ping_response, no ping_id was created, no node is added to close_nodes.
So, we need to maintain separated `last_time` field. Right?

NamsooCho: Then we can maintain precomputed key only for addable node to close_nodes.

kurnevsky: I'd suggest to use https://github.com/jeromefroe/lru-rs
We can't evict precomputed keys using time policy only - it can lead to OOM if someone decides to dos us with different keys. So we have to set limit anyway. But if we have limit and eviction strategy based on size we don't need to store time anymore.

kurnevsky: Another way we can handle it - store precomputed keys inside `DhtNode`. But this will work only for nodes that we have in our `Ktree`.

NamsooCho: I prefer `LRU' @kpp ?

kurnevsky: Or perhaps we can have global cache using https://github.com/jaemk/cached

NamsooCho: It is needed to test which case is more efficient, one case is not using precomputed keys, the other case is using `LRU` or storing precomputed key inside `DhtNode`

src/toxcore/dht/server/mod.rs
```diff
@@ -208,9 +216,21 @@ impl Server {
             lan_discovery_enabled: true,
             is_ipv6_enabled: false,
             initial_bootstrap: Vec::new(),
+            precomputed_keys: Arc::new(RwLock::new(LruCache::new(PRECOMPUTED_LRU_CACHE_SIZE))),
         }
     }
 
+    /// Get PrecomputedKey of a peer or newly make and save it.
+    fn get_precomputed_key(&self, pk: PublicKey) -> PrecomputedKey {
+        let mut precomputed_cache = self.precomputed_keys.write();
```
kpp: Better optimize it as:

1. try read lock first
2. if cache for PK has a value -> return it

3. if cache is empty
4. close read lock
5. open write lock
....

NamsooCho: https://docs.rs/lru/0.1.9/lru/struct.LruCache.html#method.get

Even `get()` requires mutable.

kpp: But https://docs.rs/lru/0.1.9/lru/struct.LruCache.html#method.contains does not.

@kurnevsky is this optimization worth implementing?

NamsooCho: contains returns `bool`.
if `contains` returns `true` then open write lock.
if `contains` returns `false` then open write lock.

kpp: if contains returns false you can compute PrecomputedKey and after that take write lock. That will reduce the time you spend in the critical section. But we'd better benchmark it.

kurnevsky: @kpp this cache takes `&mut self` on get because it tracks frequency for each element how often you get it so it can evict the less used elements. So it means that we have to acquire write lock if we want to either get or add a new key. And I think one lock is preferable because acquiring a lock should be slow operation. Minimizing instructions into critical section shouldn't worth it.

kpp: Ok

src/toxcore/net_crypto/mod.rs
```diff
@@ -173,31 +173,6 @@ impl NetCrypto {
         self.connections.read().get(&pk).cloned()
     }
 
-    /// Create `CookieResponse` packet with `Cookie` requested by `CookieRequest` packet
-    fn handle_cookie_request(&self, packet: &CookieRequest) -> Result<CookieResponse, Error> {
```
kpp: Why did you move this function from NetCrypto into DhtServer? That is not a solution.

src/toxcore/dht/server/mod.rs
```diff
@@ -1378,6 +1394,11 @@ impl Server {
     pub fn set_net_crypto(&mut self, net_crypto: NetCrypto) {
         self.net_crypto = Some(net_crypto);
     }
+
+    /// Get `PrecomputedKey`s cache.
+    pub fn get_precomputed_keys(&self) -> PrecomputedCache {
```
kpp: What's that function for?

src/toxcore/dht/precomputed_cache.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// LRU cache for `PrecomputedKey`s.
+///
+/// Calculation of `PrecomputedKey` from the `PublicKey`-`SecretKey` pair is an
+/// expensive operation. `PrecomputedKey`s should be cached whenever possible
+/// and reused later.
+#[derive(Clone)]
+pub struct PrecomputedCache {
+    sk: SecretKey,
+    precomputed_keys: Arc<RwLock<LruCache<PublicKey, PrecomputedKey>>>,
```
kpp: Maybe we should use Mutex instead of RwLock?

src/toxcore/net_crypto/mod.rs
```diff
@@ -139,6 +143,9 @@ pub struct NetCrypto {
     /// Long term keys by IP address of DHT node map. `SocketAddr` can't be used
     /// as a key since it contains additional info for `IPv6` address.
     keys_by_addr: Arc<RwLock<HashMap<(IpAddr, /*port*/ u16), PublicKey>>>,
+    /// Lru cache for precomputed keys. It stores precomputed keys to avoid
+    /// redundant calculations.
+    precomputed_keys: PrecomputedCache,
```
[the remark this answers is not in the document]
kurnevsky: What do you mean? The size is defined in dht `Server`, and the cache is shared with `net_crypto`.

src/toxcore/dht/server/mod.rs
kurnevsky: It will be used to get cache from the `Server` to use it in other modules like `net_crypto`. It's supposed to be shared. Otherwise it makes little sense to have cache in `net_crypto`.

src/toxcore/net_crypto/mod.rs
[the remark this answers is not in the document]
kpp: No it's not shared. We will have to create an instance of cache inside tox-node/src/main.rs

src/toxcore/dht/server/mod.rs
kpp: Alright. Show me usage in examples/dht.rs.

src/toxcore/net_crypto/mod.rs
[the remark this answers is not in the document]
kurnevsky: That's why I added `get_precomputed_keys` function in the `Server`. We can take the cache from server and give it to `net_crypto`.

src/toxcore/dht/server/mod.rs
kurnevsky: There is another way it could be done - just pass this cache as arg to both `Server::new` and `NetCrypto::new`. But I decided that it can be created internally in the `Server` and taken later with this method. In this way `Server::new` has less args.

kpp: Can we set the cache for net_crypto when we call DhtServer::set_net_crypto?

kurnevsky: In this case we will have to create it in `NetCrypto` and then drop it and replace with cache from `DhtServer` - it's a bit untidy.

NamsooCho:
Before LRU

before


Using LRU

after

PR #258, opened 2018-10-30
Do not remove links from the first client when the second client disconnects

kurnevsky:


src/toxcore/tcp/server/server.rs
```diff
@@ -542,6 +546,21 @@ mod tests {
         assert_eq!(packet.unwrap(), Packet::RouteResponse(
             RouteResponse { pk: client_pk_2, connection_id: 16 }
         ));
+
+        {
+            // check links
+            let state = server.state.read();
+            {
```
kurnevsky: It seems these blocks are unnecessary?

kpp: Right

kpp: Done

PR #260, opened 2018-11-09
tmp

NamsooCho:


src/toxcore/dht/packet/cookie.rs
```diff
@@ -2,17 +2,44 @@
...
+use failure::Error;
+
 use toxcore::binary_io::*;
 use toxcore::crypto_core::*;
 use toxcore::time::*;
 
 /// Number of seconds that generated cookie is valid
 pub const COOKIE_TIMEOUT: u64 = 15;
 
+/// Error that can happen when calling `get_payload` of packet.
```
kpp: That is a change from another issue)

NamsooCho: done

PR #261, opened 2018-11-09
Add Links to TcpServer for O(1) search

kpp:


src/toxcore/tcp/server/links.rs
```diff
@@ -0,0 +1,124 @@
...
+        if index < MAX_LINKS_N {
+            if let Some(mut link) = self.links[index] {
+                link.upgrade();
+            }
+        }
+    }
+
+    /** Iter over each link in self.links
+    */
+    pub fn iter_links(&self) -> Iter<Option<Link>> {
```
kurnevsky: Can we return `impl Iterator<Item = Link>` here, without option?

kpp: Yes

src/toxcore/tcp/server/server.rs
```diff
@@ -105,88 +106,120 @@ impl Server {
...
+
+        // the are both linked, send RouteResponse and
+        // send each other ConnectNotification
+        // we don't care if connect notifications fail
+        client_a.links().upgrade(b_id_in_client_a);
+        client_b.links().upgrade(a_id_in_client_b);
+
+        let client_a_notification = client_a.send_connect_notification(b_id_in_client_a + 16);
+        let client_b_notification = client_b.send_connect_notification(a_id_in_client_b + 16);
+        return Box::new(
```
kurnevsky: Clippy will be angry :)

src/toxcore/tcp/server/server.rs
```diff
@@ -105,88 +106,120 @@ impl Server {
...
-            // send RouteResponse only to current client
-            client_a.send_route_response(&packet.pk, b_id_in_client_a)
-        }
+        };
+
+        // the are both linked, send RouteResponse and
+        // send each other ConnectNotification
+        // we don't care if connect notifications fail
+        client_a.links().upgrade(b_id_in_client_a);
+        client_b.links().upgrade(a_id_in_client_b);
```
kurnevsky: It seems it might cause a deadlock if two clients send RouteRequest simultaneously and processed in parallel.

kurnevsky: But there might be another problem. What if one client sends ConnectNotification whereas another client sends DisconnectNotification? In this case links might become inconsistent - client_a will have online link but client_b will have no link.

src/toxcore/tcp/server/client.rs
```diff
@@ -60,7 +61,7 @@ impl Client {
             ip_addr,
             port,
             tx,
-            links: [None; 240],
+            links: Mutex::new(Links::new()),
```
kurnevsky: Can it be just Links, without Mutex?

kpp: No, because

```
    fn handle_route_request(&self, pk: &PublicKey, packet: &RouteRequest) -> IoFuture<()> {
        let state = self.state.read();
...
        client_a.links().upgrade(b_id_in_client_a);
        client_b.links().upgrade(a_id_in_client_b);
```

We need interior mutability but on the other side https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow_mut  :

> 

Panics if the value is currently borrowed

kpp: But we can make a single threaded tcp server with no lock at all =)

src/toxcore/tcp/server/server.rs
kpp: > In this case links might become inconsistent - client_a will have online link but client_b will have no link.

Yes there is no solution for that but to make a single threaded tcp server

src/toxcore/tcp/server/server.rs
kpp: Why?

kurnevsky: Redundant `return`.

src/toxcore/tcp/server/client.rs
kurnevsky: But it seems it can be done separately, can't it? First take client_a and update it, then release client_a and take client_b.

kpp: How would you guarantee the order?

src/toxcore/tcp/links.rs
```diff
@@ -0,0 +1,300 @@
...
+                None
+            }
+        } else {
+            None
+        }
+    }
+    /// Call Links::downgrade on the `Link` by id
+    pub fn downgrade(&mut self, index: u8) {
+        let index = index as usize;
+        assert!(index < MAX_LINKS_N);
```
kurnevsky: Maybe add a description as a second arg?

kpp: What to do mean?

kurnevsky: Like `assert!(index < MAX_LINKS_N, "The index {} is bigger than maximum allowed {}", index, MAX_LINKS_N)`.

kpp: Ah thanks

kpp: Done

src/toxcore/tcp/links.rs
```diff
@@ -0,0 +1,300 @@
...
+                Some(link)
+            } else {
+                None
+            }
+        } else {
+            None
+        }
+    }
+    /// Call Links::downgrade on the `Link` by id
+    pub fn downgrade(&mut self, index: u8) {
```
kurnevsky: How about returning `bool` if link by the index exists? It could be handy here: https://github.com/tox-rs/tox/blob/4d9ce78351e7ad78ae65b0e177b25ae1676c0306/src/toxcore/tcp/client/client.rs#L198-L206

kpp: What logic should be behind this bool?

kpp: What if the link exists and with status=Registered?

kurnevsky: This method downgrades state. So it will be always with status `Registered` if it exists. So just return true. And the same with `upgrade` method and `Online` status.

kpp: What about `try_insert`/`try_downgrade`/`try_upgrade`?

kurnevsky: You mean rename it? Because it seems no reason to have `upgrade` and `try_upgrade` at the same time.

kpp: Yes I mean rename them.

kpp: Shall I or is there any reason to do so?

kpp: Done

PR #264, opened 2018-11-12
WIP: use failure crate for DHT

NamsooCho:


.travis.yml
```diff
@@ -42,6 +42,7 @@ before_install:
   - |
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]
     then
+    echo $TRAVIS_COMMIT_RANGE
```
kpp: You don't put this changes into PR named `use failure crate for DHT`. Never.

PR #267, opened 2018-11-19
Use failure crate for DHT Packets

NamsooCho:


src/toxcore/dht/packet/cookie.rs
```diff
@@ -145,22 +145,20 @@ impl EncryptedCookie {
...
+                debug!(target: "Dht", "Cookie return deserialize error: {:?}", needed);
+                Err(GetPayloadError::IncompletePayload { packet: "Cookie".to_string(), needed, payload: self.payload.to_vec() })
             },
-            IResult::Error(e) => {
-                debug!(target: "Dht", "Cookie return deserialize error: {:?}", e);
-                Err(Error::new(ErrorKind::Other,
-                    format!("Cookie return deserialize error: {:?}", e)))
+            IResult::Error(error) => {
+                debug!(target: "Dht", "Cookie return deserialize error: {:?}", error);
+                Err(GetPayloadError::DeserializeError { packet: "Cookie".to_string(), error, payload: self.payload.to_vec() })
```
kurnevsky: Cookie is not a packet. Isn't it strange that `DeserializeError` contains `packet` field here? Do we need this field at all?

src/toxcore/dht/packet/errors.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::io::ErrorKind as IoErrorKind;
+
+/// Error that can happen when calling `get_payload` of packet.
+#[derive(Debug, Fail, PartialEq)]
+pub enum GetPayloadError {
+    /// Error indicates that received payload of encrypted packet can't be decrypted
+    #[fail(display = "Decrypt encrypted payload of {:?} packet error.", packet)]
+    DecryptError {
+        /// Packet type
+        packet: String
```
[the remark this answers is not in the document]
kurnevsky: I'd suggest to remove `packet` field. When we call `get_payload` method we know packet type statically.

src/toxcore/dht/packet/errors.rs
```diff
@@ -0,0 +1,46 @@
...
+        error: ErrorKind,
+        /// Received payload of packet
+        payload: Vec<u8>,
+    }
+}
+
+/// From trait for temporary use during transition from io:Error to custom enum error of failure crate
+impl From<GetPayloadError> for IoError {
+    fn from(_item: GetPayloadError) -> Self {
+        IoError::new(IoErrorKind::Other, "GetPayloadError occured.")
```
kurnevsky: Better add item to the text, even if it's temporary.

src/toxcore/dht/packet/errors.rs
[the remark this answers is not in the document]
kpp: Actually I agree with Namsoo. Right now the type of the Packet is included inside ADT, and keeping a special enum may cause problems, because of inconsistency. Whereas `packet: String` is for humans and pretty printing only. But we should somehow convert it to `.context`. See https://github.com/rust-lang-nursery/failure/blob/master/examples/string_custom_error_pattern.rs#L20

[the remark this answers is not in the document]
kurnevsky: Why do we need to keep the packet type inside this ADT? When we call it inside dht server we know the packet type. So we can add this info in high level error if we need it. Using string here is tempting to put things inside that have nothing to do with packets, see: https://github.com/tox-rs/tox/pull/267#discussion_r234664423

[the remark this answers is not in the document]
kpp: > Why do we need to keep the packet type inside this ADT?

I mean inside https://github.com/tox-rs/tox/blob/master/src/toxcore/dht/packet/mod.rs#L40 this ADT.

> When we call it inside dht server we know the packet type.

The question is to get the info when we are outside dht server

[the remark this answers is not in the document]
kurnevsky: So as a conclusion - we can use https://docs.rs/failure/0.1.3/failure/struct.Context.html to add a stacktrace to our errors. In this case we can check the stacktrace to see where an error appeared. So we don't need the `packet` field. Later if we need it we can add it on higher level, like:
```rust
pub enum HandlePacketError {
    ...
    GetPayloadError {
        error: GetPayloadError,
        kind: PacketKind
    },
    CookiePayloadError {
        error: GetPayloadError,
    }
    ...
}
```
Thereby, we should change `GetPayloadError` to something like:
```rust
struct GetPayloadError {
    inner: Context<GetPayloadErrorKind>
}

struct GetPayloadErrorKind {
    ... // same as current GetPayloadError, but without `packet` field
}
```

@kpp am I correct?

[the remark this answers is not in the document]
kpp: Yes

src/toxcore/dht/packet/errors.rs
NamsooCho: done

src/toxcore/dht/packet/errors.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub(crate) fn incomplete(needed: Needed, payload: Vec<u8>) -> GetPayloadError {
+        GetPayloadError::from(GetPayloadErrorKind::IncompletePayload { needed, payload })
+    }
+
+    pub(crate) fn deserialize(error: NomErrorKind, payload: Vec<u8>) -> GetPayloadError {
+        GetPayloadError::from(GetPayloadErrorKind::Deserialize { error, payload })
+    }
+}
+
+impl Fail for GetPayloadError {
```
kpp: Can `derive Fail` help us to reduce this code?

src/toxcore/dht/packet/errors.rs
```diff
@@ -0,0 +1,133 @@
...
+        error: NomErrorKind,
+        /// Received payload of packet
+        payload: Vec<u8>,
+    }
+}
+
+impl fmt::Display for GetPayloadErrorKind {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            GetPayloadErrorKind::Decrypt => {
```
kpp: Can we use `#[fail(display =` to reduce this code?

src/toxcore/dht/packet/errors.rs
NamsooCho: This test fails.
https://github.com/tox-rs/tox/blob/8a6cef4d4ec05e0f37f6ddf5d84c1c044f6a87a7/src/toxcore/dht/packet/errors.rs#L118

Don't we need `backtrace()`?

kpp: I thought this methods is autogenerated with `derive Fail`.

PR #269, opened 2018-11-23
chore(ci): improve TravisCI

kpp:
Speed up TravisCI:

before: Ran for 33 min 4 sec, Total time 1 hr 42 min 25 sec
after: Ran for 23 min 10 sec Total time 1 hr 18 min 6 sec

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,12 @@
 language: rust
+os:
+  - linux
+  - osx
 rust:
   - 1.26.0
   - stable
   - beta
   - nightly
-cache:
-  - apt
-  - cargo
-# needed for `cargo install cargo-travis`
-addons:
-  apt:
-    packages:
-      - libcurl4-openssl-dev
-      - libelf-dev
-      - libdw-dev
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -42,38 +20,57 @@ before_install:
...
+            - libelf-dev
+            - libdw-dev
+            - binutils-dev
+            - cmake
+          sources:
+            - kalakris-cmake
+      cache:
+        directories:
+          - /home/travis/.cargo
+          - /home/travis/build/tox-rs/tox/target/kcov-master
```
kurnevsky: `cache cargo` does almost the same: https://docs.travis-ci.com/user/caching/#rust-cargo-cache
Why cache only `kcov-master` inside `target`?

kpp: https://levans.fr/rust_travis_cache.html

kpp: > Why cache only kcov-master inside target?

Because we don't need to cache target/debug because it will be rebuilt for sure. So it's a waste of time and the resources of travis

PR #271, opened 2018-11-27
WIP: congestion control implementation

kurnevsky:


src/toxcore/net_crypto/crypto_connection.rs
```diff
@@ -9,9 +9,8 @@ use toxcore::crypto_core::*;
 use toxcore::dht::packet::*;
 use toxcore::time::*;
 
-/// How often in seconds `CookieRequest` or `CryptoHandshake` packets should be
-/// sent
-pub const CRYPTO_SEND_PACKET_INTERVAL: u64 = 1;
+/// Interval in ms between sending cookie request/handshake packets.
+pub const CRYPTO_SEND_PACKET_INTERVAL: u64 = 1000;
```
kpp: why?

kurnevsky: Because it's used as milliseconds more often: https://github.com/tox-rs/tox/pull/271/files#diff-dd71e0765ce00efdcbe43457539f6e18R565 https://github.com/tox-rs/tox/pull/271/files#diff-dd71e0765ce00efdcbe43457539f6e18R569

And I had a bug there when forgot to multiply it by 1000. So I decided that having it in ms is clearer.

src/toxcore/net_crypto/crypto_connection.rs
```diff
@@ -27,6 +26,34 @@ pub const DEFAULT_RTT: u64 = 1000;
...
+/// How many last sizes of `send_array` should be recorded for congestion
+/// control.
+pub const CONGESTION_QUEUE_ARRAY_SIZE: usize = 12;
+
+/// How many numbers of sent lossless packets should be recorded for congestion
+/// control. It should be bigger than `CONGESTION_QUEUE_ARRAY_SIZE` due to rtt.
+pub const CONGESTION_LAST_SENT_ARRAY_SIZE: usize = CONGESTION_QUEUE_ARRAY_SIZE * 2;
+
+/// Minimum packets rate per second.
+pub const CRYPTO_PACKET_MIN_RATE: f64 = 4.0;
```
NamsooCho: Why f64? Not f32. Does it need such large precision?

kurnevsky: Why would I choose f32? Is it performance or memory? I think the performance difference should be negligible, we can check later. Right now I just don't want to care about the precision so I use the higher type.

src/toxcore/net_crypto/packets_array.rs
```diff
@@ -119,6 +119,17 @@ impl<T> PacketsArray<T> {
...
 
+    /// Check if packet the index exists
+    pub fn contains(&self, index: u32) -> bool {
+        let len = self.len();
+
+        if self.buffer_end.overflowing_sub(index).0 > len || index.overflowing_sub(self.buffer_start).0 >= len {
+            return false
+        }
+
+        self.buffer[real_index(index)].is_some()
```
kpp: O(n)?

kurnevsky: It's O(1). Where do you see O(n)?

PR #277, opened 2018-12-04
Use bounded streams

kurnevsky:


src/toxcore/tcp/server/server.rs
```diff
@@ -375,11 +375,10 @@ impl Server {
             .collect::<Vec<PublicKey>>();
 
         let remove_timedouts = keys.iter()
-            .map(|key| {
-                self.shutdown_client_inner(key, state)
-            });
+            // failure in removing one client should not affect other clients
+            .map(|key| self.shutdown_client_inner(key, state).then(|_| Ok(())));
```
kpp: What's that for?

kurnevsky: `shutdown_client_inner` may return failed future if it can't send disconnect notification. In this case we should ignore it and send all other notifications that can be sent.

kpp: OK, thanks

PR #280, opened 2018-12-11
TCP client links

kurnevsky:


src/toxcore/tcp/client/client.rs
```diff
@@ -353,24 +318,16 @@ impl Client {
...
         // it is important that the result future succeeds only if packet is
         // sent since we take only one successful future from several relays
         // when send data packet
-        let pk_to_id = self.pk_to_id.read();
-        if let Some(&connection_id) = pk_to_id.get(&destination_pk) {
-            let links = self.links.read();
-            if links[connection_id as usize - 16].map(|link| link.status) == Some(LinkStatus::Online) {
+        let links = self.links.read();
+        if let Some(index) = links.id_by_pk(&destination_pk) {
+            if links.by_id(index).map(|link| link.status) == Some(LinkStatus::Online) {
```
kpp: How about add a new `fn by_pk`?

kurnevsky: I also thought about it but then realized that index is also needed to send it inside Data packet. It means that this function will have to return tuple with Link and its index. So I decided that I can just call 2 different functions. Do you think that it worth to add such function?

kpp: > Do you think that it worth to add such function?

No

PR #281, opened 2018-12-13
Add crypto_init() to all tests which use sodium

### LINKED ISSUE #279: Sometimes, cargo test fails on Mac OSX
```rust
failures:

---- toxcore::crypto_core::tests::encrypt_data_symmetric_test stdout ----
thread 'toxcore::crypto_core::tests::encrypt_data_symmetric_test' panicked at 'called `Result::unwrap()` on an `Err` value: ()', libcore/result.rs:1009:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    toxcore::crypto_core::tests::encrypt_data_symmetric_test

test result: FAILED. 527 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

NamsooCho:
Close #279

### DIFF
--- a/examples/dht_server.rs
+++ b/examples/dht_server.rs
@@ -58,7 +58,7 @@ fn bind_socket(addr: SocketAddr) -> UdpSocket {
 fn main() {
     env_logger::init();
 
-    if !crypto_init() {
+    if crypto_init().is_err() {
         panic!("Crypto initialization failed.");
     }
 
--- a/src/toxcore/binary_io.rs
+++ b/src/toxcore/binary_io.rs
@@ -101,6 +101,11 @@ macro_rules! encode_decode_test (
     ($test:ident, $value:expr) => (
         #[test]
         fn $test() {
+            #[cfg(test)]
+            use toxcore::crypto_core::*;
+
+            crypto_init().unwrap();
+
             let value = $value;
             let mut buf = [0; 1024 * 1024];
             let (_, size) = value.to_bytes((&mut buf, 0)).unwrap();
--- a/src/toxcore/crypto_core.rs
+++ b/src/toxcore/crypto_core.rs
@@ -15,20 +15,20 @@ use toxcore::binary_io::*;
 
 Runs [`sodiumoxide::init()`](../../../sodiumoxide/fn.init.html).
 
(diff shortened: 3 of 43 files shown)


kpp:
Does it help to `cargo test -- --test-threads=4`?

NamsooCho:
Yes. I am using 4 core, 8 threads CPU.

kpp:
How many times did you run `cargo test`?

kpp:
diff stat: `+231 -0`. How is it possible? We got ~580` tests and you changed only 231?

NamsooCho:
I didn't add `crypt_init()` to test function which is not using sodium.
And I run it 40 times with 8 threads.

NamsooCho:
For example,
https://github.com/tox-rs/tox/blob/738c9598f63ea51ac8ca19de00d8778d2a0ed15f/src/toxcore/dht/server/mod.rs#L1427

this fn is called by almost every test fn.

kurnevsky:
What about toxencryptsave tests?

kpp:
From sodiumoxide: `pub fn init() -> Result<(), ()>`, maybe we should change the API to return Result and to be warned about unused result? Then we can use `crypto_init().unwrap()`

kurnevsky:
@kpp do we need to wrap `sodiumoxide::init()` in this case?

kpp:
Why not?

kurnevsky:
What purpose does it serve?

kpp:
You don’t call a function from external crate to make it work

kurnevsky:
> You don’t call a function from external crate to make it work

It would make sense if we had wrappers for all sodiumoxide functions and types. But we call sodiumoxide directly from a lot of places (`gen_key()`, `gen_nonce()`, etc) and use its types (PublicKey, Nonce, etc). So wrapping only `init()` seems useless to me - it can't simplify anything, but we have to remember about this custom wrapper and what it does.

kpp:
> But we call sodiumoxide directly from a lot of places

Yes but we don't call it from tox-node, which is a final product.

kurnevsky:
> Yes but we don't call it from tox-node, which is a final product.

ok then, let it be wrapped.

PR #284, opened 2018-12-14
Improve README + revert gpl license

kpp:


quininer:
@kpp My contribution is really insignificant. no need to include my name.

PR #285, opened 2018-12-15
Use failure crate in lan_discovery with Error-ErrorKind pattern

NamsooCho:


examples/dht_server.rs
```diff
@@ -6,10 +6,10 @@ extern crate futures;
 extern crate tokio;
 extern crate hex;
 
-#[macro_use]
+//#[macro_use]
 extern crate log;
 extern crate env_logger;
-
+/*
```
kurnevsky: Is it WIP?

NamsooCho: Opps. done

src/toxcore/io_tokio.rs
```diff
@@ -1,17 +1,91 @@
...
-pub type IoStream<T> = Box<Stream<Item = T, Error = Error> + Send>;
+pub type IoStream<T> = Box<Stream<Item = T, Error = IoError> + Send>;
+
+/// Error that can happen during sending all packet to peer.
+#[derive(Debug)]
+pub struct SendAllToError {
+    ctx: Context<SendAllToErrorKind>,
+}
+
+impl SendAllToError {
```
kurnevsky: Shouldn't it be named just `SendToError`? It seems `send_to` function will have the same error.
Also `send_to` function can't return `TimedOut` error. Maybe split error types for `send_to` and `send_to_bounded`?

examples/dht_server.rs
```diff
@@ -56,6 +58,7 @@ fn bind_socket(addr: SocketAddr) -> UdpSocket {
 }
 
 fn main() {
+
```
kurnevsky: Unnecessary change.

src/toxcore/io_tokio.rs
NamsooCho: I think `TimedOut` error occurs during sending packet(s).
So, `SendToError` can include `TimedOut`.

examples/dht_server.rs
NamsooCho: done

src/toxcore/io_tokio.rs
kurnevsky: No, it can occur only if we set timeout ourselves.

src/toxcore/io_tokio.rs
```diff
@@ -1,17 +1,91 @@
...
+
+/// The specific kind of error that can occur.
+#[derive(Debug, Fail)]
+pub enum SendToErrorKind {
+    /// Send error
+    #[fail(display = "Send packet error: {:?}", error)]
+    Eof {
+        /// Send all packet error
+        #[fail(cause)]
+        error: IoError
```
kurnevsky: No need to include error here - we don't have it.

src/toxcore/io_tokio.rs
```diff
@@ -1,17 +1,91 @@
...
+    }
+}
+
+/// The specific kind of error that can occur.
+#[derive(Debug, Fail)]
+pub enum SendToErrorKind {
+    /// Send error
+    #[fail(display = "Send packet error: {:?}", error)]
+    Eof {
+        /// Send all packet error
```
kurnevsky: Just `Send packet error`, without `all`.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -128,11 +196,11 @@ impl LanDiscoverySender {
 
     /// Run LAN discovery periodically. Result future will never be completed
     /// successfully.
-    pub fn run(mut self) -> IoFuture<()> {
+    pub fn run(mut self) -> impl Future<Item=(), Error=LanDiscoveryError> + Send {
```
kurnevsky: Do we need to wrap errors when there is only one possible error (`tokio::timer::Error` here)? @kpp ?

src/toxcore/io_tokio.rs
```diff
@@ -36,37 +106,36 @@ pub fn send_to_bounded<T: Send + 'static, Tx, E: Debug>(tx: &Tx, v: T, timeout:
...
         .timeout(timeout)
-        .map_err(|e| Error::new(ErrorKind::Other,
+        .map_err(|e| IoError::new(IoErrorKind::Other,
             format!("Failed to send message: {}", e)
         )))
 }
 
 /// Send item to a sink using reference
-pub fn send_all_to<T: Send + 'static, S, Tx, E: Debug>(tx: &Tx, s: S) -> IoFuture<()>
+pub fn send_all_to<T: Send + 'static, S, Tx, E: Debug>(tx: &Tx, s: S) -> impl Future<Item=(), Error=SendToError> + Send
```
kurnevsky: Actually I think we don't need to wrap errors for `send_to*` methods. They can return only error that is defined in their signature as `SinkError` and also they are supposed for internal usage only. So I'd only change it's return type to `impl Future<Item=(), Error=E> + Send` and wrap it only on higher level. @kpp do you agree?

kpp: Then we should mark them as `pub(crate)`. Why not Error=SinkError?

kurnevsky: ok, agree. Later when we get rid of `IoFuture` we can just make the whole module `binary_io` private.

> Why not Error=SinkError?

Because `SinkError` is a trait, E is generic parameter `E: SinkError`.

kpp: > So I'd only change it's return type to impl Future<Item=(), Error=E> + Send and wrap it only on higher level.

How would you wrap it?

kurnevsky: We'll need to wrap it to another `ErrorKind` anyway whether it's our custom error or tokio error. It seems for each module (tcp, dht, lan discovery, etc) it will be it's own ErrorKind.

kurnevsky: Also I'd suggest to try this approach first before we decide. I'm not sure how convenient it will be.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -1,20 +1,102 @@
...
+
+/// The specific kind of error that can occur.
+#[derive(Debug, Fail)]
+pub enum LanDiscoveryErrorKind {
+    /// Ping wakeup timer error
+    #[fail(display = "Lan discovery wakeup timer error")]
+    Wakeup,
+    /// Send packet(s) error
+    #[fail(display = "Send packet(s) error")]
+    SendTo,
```
kurnevsky: There shouldn't be such error. It's possible that we can't send a packet due to busy socket and timeout. In this case we don't want to stop node/client.

kurnevsky: Or at least it shouldn't be returned by `LanDiscovery::run`. Which means we'd probably need separate errors for `LanDiscovery::run` and `LanDiscovery::send`.

NamsooCho: done

src/toxcore/dht/lan_discovery.rs
```diff
@@ -124,15 +206,16 @@ impl LanDiscoverySender {
         );
 
         send_all_to_bounded(&self.tx, stream, Duration::from_secs(LAN_DISCOVERY_SEND_TIMEOUT))
+            .map_err(|e| LanDiscoveryError::from(e))
```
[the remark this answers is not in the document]
kurnevsky: But we don't return this error, we only log it. And we explicitly write it with text:
```rust
warn!("Failed to send LAN discovery packets: {}", e);
```
So it doesn't make sense to wrap it to indicate the place.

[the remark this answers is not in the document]
NamsooCho: done

src/toxcore/dht/lan_discovery.rs
```diff
@@ -1,20 +1,89 @@
...
+use std::io::Error as IoError;
+use std::io::ErrorKind as IoErrorKind;
+
+use failure::{Backtrace, Context, Fail};
 use futures::{future, stream, Future, Stream};
 use futures::sync::mpsc;
 use get_if_addrs;
 use get_if_addrs::IfAddr;
 use tokio::timer::Interval;
+use tokio::timer::Error as WakeupError;
```
kpp: WakeupError -> TimerError because Wakeup is a "keyword" for Futures

src/toxcore/dht/lan_discovery.rs
```diff
@@ -128,21 +197,21 @@ impl LanDiscoverySender {
...
         let wakeups = Interval::new(Instant::now(), interval);
-        let future = wakeups
-            .map_err(|e| Error::new(ErrorKind::Other, format!("LanDiscovery timer error: {:?}", e)))
+        wakeups
+            .map_err(|e| LanDiscoveryError::from(e))
             .for_each(move |_instant| {
                 trace!("LAN discovery sender wake up");
                 self.send().then(|res| {
                     if let Err(e) = res {
                         warn!("Failed to send LAN discovery packets: {}", e);
```
kurnevsky: This error should be returned since you already handled timeouts.

src/toxcore/dht/server/mod.rs
```diff
@@ -616,7 +616,8 @@ impl Server {
             (packet, addr)
         }).collect::<Vec<_>>();
 
-        send_all_to_bounded(&self.tx, stream::iter_ok(packets), Duration::from_secs(DHT_SEND_TIMEOUT))
+        Box::new(send_all_to_bounded(&self.tx, stream::iter_ok(packets), Duration::from_secs(DHT_SEND_TIMEOUT))
```
kpp: Why Box?

src/toxcore/io_tokio.rs
```diff
@@ -36,37 +36,40 @@ pub fn send_to_bounded<T: Send + 'static, Tx, E: Debug>(tx: &Tx, v: T, timeout:
...
     where S: Stream<Item = T, Error = E> + Send + 'static,
           Tx: Sink<SinkItem = T, SinkError = E> + Send + Clone + 'static
 {
-    Box::new(send_all_to(tx, s)
+    send_all_to(tx, s)
         .timeout(timeout)
-        .map_err(|e| Error::new(ErrorKind::Other,
-            format!("Failed to send message: {}", e)
-        )))
+        .then(|r| {
```
kurnevsky: Maybe handle timeout error in `LanDiscovery` module? This way we can log it properly.

src/toxcore/dht/server/mod.rs
NamsooCho: https://github.com/tox-rs/tox/blob/65174889f4301fcca24da059fd436b9f5558ee59/src/toxcore/dht/server/mod.rs#L603
Because return type of function is still IoFuture<()>. It is temporary used. After we change all IoFuture<()> to impl Future, this Box can be removed.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -1,20 +1,100 @@
...
+#[derive(Debug, Fail)]
+pub enum LanDiscoveryErrorKind {
+    /// Ping wakeup timer error
+    #[fail(display = "Lan discovery wakeup timer error")]
+    Wakeup,
+    /// Send packet(s) error
+    #[fail(display = "Send packet(s) error: {:?}", error)]
+    SendTo {
+        /// IO error during sending packet(s).
+        error: IoError,
```
kurnevsky: Let's pack it into `Context`.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -128,21 +208,24 @@ impl LanDiscoverySender {
...
+            .map_err(|e| LanDiscoveryError::from(e))
             .for_each(move |_instant| {
                 trace!("LAN discovery sender wake up");
-                self.send().then(|res| {
-                    if let Err(e) = res {
+                self.send().then(|r| {
+                    if let Err(e) = r {
                         warn!("Failed to send LAN discovery packets: {}", e);
+                        if e.is_inner() {
+                            let io_error = e.into_inner().expect("Missing inner error.");
```
kurnevsky: `if let Some(...) = ...`

NamsooCho: done

src/toxcore/dht/lan_discovery.rs
```diff
@@ -1,20 +1,101 @@
...
+}
+
+impl fmt::Display for LanDiscoveryError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.ctx.fmt(f)
+    }
+}
+
+/// The specific kind of error that can occur.
+#[derive(Debug, Fail)]
```
kurnevsky: We can implement `PartialEq` now. To be consistent with other error enums.

src/toxcore/dht/lan_discovery.rs
```diff
@@ -128,21 +209,25 @@ impl LanDiscoverySender {
...
+        wakeups
+            .map_err(|e| LanDiscoveryError::from(e))
             .for_each(move |_instant| {
                 trace!("LAN discovery sender wake up");
-                self.send().then(|res| {
-                    if let Err(e) = res {
+                self.send().then(|r| {
+                    if let Err(e) = r {
                         warn!("Failed to send LAN discovery packets: {}", e);
+                        if e.is_inner() {
```
kurnevsky: No need to check `is_inner` since you check it with `if let` construction.

NamsooCho: done

PR #286, opened 2018-12-19
Handle TCP reconnections correctly

kurnevsky:


src/toxcore/tcp/server/server.rs
```diff
@@ -54,14 +54,24 @@ impl Server {
...
+        let shutdown_future = if state.connected_clients.contains_key(&client.pk()) {
+            self.shutdown_client_inner(&client.pk(), &mut state)
+        } else {
+            Box::new(future::ok(()))
+        };
+
         state.keys_by_addr
             .insert((client.ip_addr(), client.port()), client.pk());
         state.connected_clients
             .insert(client.pk(), client);
```
kpp: > state.connected_clients.insert

And you will replace the existing client without disconnecting it.

kurnevsky: Nope, I called `shutdown_client_inner` for it. So there won't be a client with the same PK anymore.

kpp: You are right

PR #287, opened 2018-12-19
Add ConnectionId as separate type

kurnevsky:


src/toxcore/tcp/codec.rs
```diff
@@ -361,29 +362,14 @@ mod tests {
         // not enough bytes to decode Packet
         assert!(bob_codec.decode(&mut buf).unwrap().is_none());
     }
-    #[test]
-    fn decode_packet_error() {
```
kpp: Why did you remove this test?

src/toxcore/tcp/connection_id.rs
```diff
@@ -0,0 +1,92 @@
...
+use toxcore::tcp::links::MAX_LINKS_N;
+
+use nom::be_u8;
+
+/// Connection ID is either a number between [16, 255] or 0. Zero can be
+/// included in a response and means that the previous request was invalid.
+/// If a connection id is a number between [16, 255] it can be uniquely mapped
+/// to a connection index between [0, 239].
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct ConnectionId(u8);
```
kpp: How about `ConnectionId(Option<u8>)`?

src/toxcore/tcp/codec.rs
kurnevsky: Because it's broken. It meant to test deserialization with connection id = 0 but it doesn't. It fails on decryption step due to wrong codec with wrong key being used. Also, connection id = 0 is valid case now.

src/toxcore/tcp/connection_id.rs
kurnevsky: It's internal private state so won't change a lot. But it will be less memory efficient because Option<u8> can't be optimized to single u8.

kurnevsky: We can use Option<NonZeroU8>: https://doc.rust-lang.org/std/num/struct.NonZeroU8.html . But it will change things only inside `ConnectionId` so should we?

kpp: Oh thanks! `Option<NonZeroU8>` is very idiomatic in our case.

src/toxcore/tcp/codec.rs
kpp: But you can test it with id=1

kurnevsky: ok, rewrote this test completely with id = 0x0F (1 is also valid packet id).

src/toxcore/tcp/connection_id.rs
kurnevsky: Check the last commit. `Option<NonZeroU8>` only complicates the code and adds one more `unwrap`. Do you still think we need it?

kpp: The API is the same but the inner representation is more idiomatic. I think we should use it.

PR #288, opened 2018-12-21
Fix clippy warnings

kurnevsky:


src/toxcore/state_format/old.rs
```diff
@@ -525,6 +526,7 @@ impl FromBytes for FriendState {
 }
 
 impl ToBytes for FriendState {
+    #[allow(clippy::cyclomatic_complexity)]
```
kpp: That's not a fix)

kurnevsky: But it's a false positive warning due to macro expansion. By definition, cyclomatic complexity is a quantitative measure of the number of linearly independent paths through a program's source code. Obviously, this this method has only one path, which means it's cyclomatic complexity is 1.

src/toxcore/onion/packet/mod.rs
```diff
@@ -258,7 +258,7 @@ impl ToBytes for OnionReturn {
 }
 
 impl OnionReturn {
-    #[cfg_attr(feature = "cargo-clippy", allow(clippy::needless_pass_by_value))]
+    #[allow(clippy::needless_pass_by_value)]
```
kpp: Why needless_pass_by_value? Can you change inner to `&Option<OnionReturn>` using Option::as_ref in the call site?

kurnevsky: It's not the Option that gives this warning, it's `buf: (&'a mut [u8], usize)`. Clippy suggest to pass it by ref as `buf: &(&'a mut [u8], usize)` which is not very idiomatic for `cookie-factory` (however doable). Shall I change it?

kpp: Ah I see. No, keep it.

PR #289, opened 2018-12-22
Apply Error-ErrorKind pattern to DHT Codec

NamsooCho:


src/toxcore/dht/codec.rs
```diff
@@ -1,24 +1,81 @@
...
+        DecodeError { ctx }
+    }
+}
+
+/// Error that can happen when decoding `Packet` from bytes
+#[derive(Clone, Debug, Eq, PartialEq, Fail)]
+pub enum DecodeErrorKind {
     /// Error indicates that we received too big packet
-    #[fail(display = "Packet should not be longer then 2048 bytes: {} bytes", len)]
+    #[fail(display = "Packet should not be longer then 2048 bytes: {} bytes longer", len)]
```
kpp: Remove the last `longer` and change then -> than

src/toxcore/dht/codec.rs
```diff
@@ -33,22 +90,95 @@ pub enum DecodeError {
...
-    DeserializeError {
+    Deserialize {
         /// Parsing error
         error: ErrorKind,
         /// Received packet
         packet: Vec<u8>,
+    },
+    /// For From<DecodeError> trait.
+    #[fail(display = "Error: {}", msg)]
+    String {
```
kpp: String -> Custom ?

kpp: Also why don't you keep IoError inside the enum?

src/toxcore/dht/codec.rs
```diff
@@ -33,22 +90,95 @@ pub enum DecodeError {
...
+#[derive(Debug, PartialEq, Fail)]
+pub enum EncodeErrorKind {
     /// Error indicates that `Packet` is invalid and can't be serialized
     #[fail(display = "Serialize Packet error: {:?}", error)]
-    SerializeError {
+    Serialize {
         /// Serialization error
         error: GenError
+    },
+    /// For From<ENcodeError> trait.
```
kpp: ENcodeError?

src/toxcore/dht/server_ext.rs
```diff
@@ -34,7 +35,12 @@ impl ServerExt for Server {
                 Err(ref e) => {
                     error!("packet receive error = {:?}", e);
                     // ignore packet decode errors
-                    e.as_fail().downcast_ref::<DecodeError>().is_none()
+                    let fail: &Fail = e;
+                    if let Some(cause) = fail.cause() {
+                        cause.downcast_ref::<DecodeError>().is_none()
```
kurnevsky: You shouldn't need downcast here anymore, you can just check the error kind. The logic behind this check is that if it's packet decode error we continue the processing, but if it's `IoError` thrown by tokio we terminate the processing.

kpp: @kurnevsky how would you rewrite it?

kurnevsky: @kpp first your previous remarks should be fixed. Then we can just check if the kind is whatever  will be instead of `EncodeErrorKind::String`. I can show later if @NamsooCho has troubles.

src/toxcore/dht/codec.rs
kurnevsky: It's for error from tokio, maybe just call it `IoError`? `Custom` is not very explanatory.

src/toxcore/dht/codec.rs
```diff
@@ -33,22 +90,95 @@ pub enum DecodeError {
...
-    DeserializeError {
+    Deserialize {
         /// Parsing error
         error: ErrorKind,
         /// Received packet
         packet: Vec<u8>,
+    },
+    /// For From<DecodeError> trait.
+    #[fail(display = "Error: {}", msg)]
+    IoError {
```
kurnevsky: Maybe just `Io` to make it consistent with `Deserialize`?

src/toxcore/dht/codec.rs
```diff
@@ -33,22 +90,95 @@ pub enum DecodeError {
...
+        /// Error message
+        msg: String,
```
kurnevsky: IoError should be stored here instead of string.

src/toxcore/dht/codec.rs
```diff
@@ -33,22 +90,95 @@ pub enum DecodeError {
...
-    SerializeError {
+    Serialize {
         /// Serialization error
         error: GenError
+    },
+    /// For From<EncodeError> trait.
+    #[fail(display = "Error: {}", msg)]
+    IoError {
+        /// Error message
+        msg: String,
```
kurnevsky: IoError should be stored here instead of string.

src/toxcore/dht/codec.rs
```diff
@@ -33,22 +90,95 @@ pub enum DecodeError {
...
     /// Error indicates that `Packet` is invalid and can't be serialized
     #[fail(display = "Serialize Packet error: {:?}", error)]
-    SerializeError {
+    Serialize {
         /// Serialization error
         error: GenError
+    },
+    /// For From<EncodeError> trait.
+    #[fail(display = "Error: {}", msg)]
+    IoError {
```
kurnevsky: Maybe just `Io` to make it consistent with `Serialize`?

NamsooCho: done

src/toxcore/dht/server_ext.rs
```diff
@@ -6,9 +6,10 @@ use std::net::{SocketAddr, IpAddr};
 use futures::{future, Future, Sink, Stream};
 use futures::sync::mpsc::Receiver;
 use tokio::net::{UdpSocket, UdpFramed};
+use failure::Fail;
```
kurnevsky: No need in this anymore?

src/toxcore/dht/codec.rs
```diff
@@ -290,8 +423,10 @@ mod tests {
         let mut buf = BytesMut::new();
 
         buf.extend_from_slice(b"\xFF");
+
+        let res = codec.decode(&mut buf);
         // not enought bytes to decode EncryptedPacket
-        assert!(codec.decode(&mut buf).is_err());
+        assert!(res.is_err());
```
kurnevsky: Check for exact error?

NamsooCho: Can't derive PartialEq bacause std::io::Error doesn't implement PartialEq

src/toxcore/dht/server_ext.rs
NamsooCho:
```
error[E0599]: no method named `compat` found for type `toxcore::dht::codec::DecodeError` in the current scope
  --> src/toxcore/dht/server_ext.rs:51:55
   |
51 |         }).map_err(|e| Error::new(ErrorKind::Other, e.compat()));
   |                                                       ^^^^^^
   | 
```

kpp:
Looks good to me. @kurnevsky ?

src/toxcore/dht/codec.rs
NamsooCho: PartialEq is added.

src/toxcore/dht/codec.rs
```diff
@@ -33,23 +90,90 @@ pub enum DecodeError {
...
+pub enum EncodeErrorKind {
     /// Error indicates that `Packet` is invalid and can't be serialized
     #[fail(display = "Serialize Packet error: {:?}", error)]
-    SerializeError {
+    Serialize {
         /// Serialization error
         error: GenError
-    }
+    },
+    /// For From<EncodeError> trait.
```
kurnevsky: It's for From<IoError> for EncodeError, but I'd write something like:
```rust
/// General IO error that can happen with UDP socket.
```
Otherwise looks good.

NamsooCho: done

PR #299, opened 2019-01-15
Add timeout for the TCP handshake

kurnevsky:


src/toxcore/tcp/server/server_ext.rs
```diff
@@ -23,6 +25,9 @@ use crate::toxcore::stats::*;
 /// Interval in seconds for Tcp Ping sender
 const TCP_PING_INTERVAL: u64 = 1;
 
+/// Timeout in seconds for the TCP handshake.
+const TCP_HANDSHAKE_TIMEOUT: u64 = 10;
```
kpp: isn't 10 seconds too much?

kurnevsky: I took it from here: https://github.com/TokTok/c-toxcore/blob/98b754e1f22220f10196760389371dba6d139c6f/toxcore/TCP_client.h#L30

PR #300, opened 2019-01-16
Remove useless logs

kurnevsky:
These logs duplicate log messages from `ServerExt` and not so detailed.

### DIFF
--- a/src/toxcore/dht/server/mod.rs
+++ b/src/toxcore/dht/server/mod.rs
@@ -626,102 +626,39 @@ impl Server {
     /// Function to handle incoming packets and send responses if necessary.
     pub fn handle_packet(&self, packet: Packet, addr: SocketAddr) -> impl Future<Item = (), Error = Error> + Send {
         match packet {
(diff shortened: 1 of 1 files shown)


src/toxcore/dht/server/mod.rs
```diff
@@ -626,102 +626,39 @@ impl Server {
     /// Function to handle incoming packets and send responses if necessary.
     pub fn handle_packet(&self, packet: Packet, addr: SocketAddr) -> impl Future<Item = (), Error = Error> + Send {
         match packet {
-            Packet::PingRequest(packet) => {
-                debug!("Received ping request");
```
NamsooCho: Why is this useless?

kurnevsky: As I wrote, it duplicates the log message here: https://github.com/tox-rs/tox/blob/master/src/toxcore/dht/server_ext.rs#L43

NamsooCho: Ah Thanks!

PR #305, opened 2019-01-21
Add standalone node example

povilasb:
Adds a new example binary with which you can build a totally separate p2p network.
This is convenient for debugging, trying out new ideas or just experimenting with the library API. 
This example is different from `dht_server` in a way that it doesn't rely on any existing nodes to bootstrap to the network.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,9 @@ failure = "0.1"
 lru = "0.1"
 
 [dev-dependencies]
+clap = "2.32.0"
 env_logger = "0.6"
 hex = "0.3"
 tokio-timer = "0.2"
 tokio-executor = "0.1"
+unwrap = "1.2.1"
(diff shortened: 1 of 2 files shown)


kpp:
@povilasb Hey! Thanks for your PR.

Do you think this is a example whereas there is a binary crate with clap-rs args, config-rs and with possibility to launch either a dht or a tcp node or both of them? See https://github.com/tox-rs/tox-node

We should definitely add a link to tox-rs/tox-node to README.md

povilasb:
Cool, thanks :)
In this case probably this example is redundant cause it does not demonstrate more API than the existing `dht_server` example. So closing.

kpp:
> We should definitely add a link to tox-rs/tox-node to README.md

Added to https://github.com/tox-rs/tox#progress

PR #307, opened 2019-01-29
Use crate failure on DHT server

NamsooCho:


src/toxcore/dht/packet/mod.rs
```diff
@@ -17,7 +17,7 @@ mod lan_discovery;
 mod crypto_handshake;
 mod crypto_data;
 mod cookie;
-mod errors;
+pub mod errors;
```
kurnevsky: Why `pub mod` when you have `pub use` next?

NamsooCho: https://github.com/tox-rs/tox/blob/25db53d1eb30b2e0a49d4b947a2f3138d5770ce1/src/toxcore/onion/packet/onion_announce_request.rs#L8

errors.rs is used by onion module.

src/toxcore/dht/server_ext.rs
```diff
@@ -66,7 +66,8 @@ impl ServerExt for Server {
             .select(network_writer)
             .map(|_| ())
             .map_err(|(e, _)| e)
-            .select(self.run())
+            .select(self.run()
+                .map_err(|e| Error::new(ErrorKind::Other, format!("{:?}", e))))
```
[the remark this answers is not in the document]
NamsooCho: done

src/toxcore/dht/server/mod.rs
```diff
@@ -530,10 +549,20 @@ impl Server {
             &payload
         ));
         self.send_to_direct(node.saddr, ping_req)
+            .then(|r| {
```
kurnevsky: I don't like this block added multiple times to sending packet methods. A fortiori, we didn't filter out timeout error before. So I'd add this error to the return type: `Error = TimeoutError<mpsc::SendError<(Packet, SocketAddr)>>`.

src/toxcore/dht/server/mod.rs
```diff
@@ -640,31 +699,33 @@ impl Server {
             Packet::OnionRequest2(packet) => Box::new(self.handle_onion_request_2(&packet, addr)),
             Packet::OnionAnnounceRequest(packet) => Box::new(self.handle_onion_announce_request(packet, addr)),
             Packet::OnionDataRequest(packet) => Box::new(self.handle_onion_data_request(packet)),
-            Packet::OnionResponse3(packet) => Box::new(self.handle_onion_response_3(packet)),
-            Packet::OnionResponse2(packet) => Box::new(self.handle_onion_response_2(packet)),
-            Packet::OnionResponse1(packet) => Box::new(self.handle_onion_response_1(packet)),
-            Packet::BootstrapInfo(packet) => Box::new(self.handle_bootstrap_info(&packet, addr)),
+            Packet::OnionResponse3(packet) => Box::new(self.handle_onion_response_3(packet)
+                .map_err(|e| HandlePacketError::from(e))),
```
kurnevsky: Perhaps move these `map`s to `handle_onion_response_*` to be consistent with other handling packets methods?

kurnevsky:
This PR adds many similar errors: `BootstrapRequestsErrorKind { Wakeup, SendTo }, DhtLoopErrorKind { Wakeup, SendTo }, OnionKeyRefreshingErrorKind { Wakeup, SendTo }, PingsSendingErrorKind { Wakeup, SendTo }`. Do we really need them all? Maybe add one error `RunError` and use it for all `run_*` methods?

kurnevsky:
Also I'm thinking that we should eliminate `send_to_bounded` and `send_all_to_bounded` functions. Instead, we can explicitly add single timeout in each `run_*` function and handle it properly. But it can be done later.

NamsooCho:
>This PR adds many similar errors: ...

I worried that we can't distinguish where is the error if we use one error-type.

src/toxcore/dht/server/mod.rs
```diff
@@ -370,11 +384,11 @@ impl Server {
 
     /// Run DHT main loop periodically. Result future will never be completed
     /// successfully.
-    fn run_main_loop(self) -> impl Future<Item = (), Error = Error> + Send {
+    fn run_main_loop(self) -> impl Future<Item = (), Error = MainLoopError> + Send {
```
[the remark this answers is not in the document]

kurnevsky:
> I worried that we can't distinguish where is the error if we use one error-type.

We discussed a very similar situation before with `GetPayloadError` and decided that we can use backtraces to find the exact place of the error. It seems it can be applicable here as well?

NamsooCho:
Yes. It was just worthless worry.

PR #309, opened 2019-01-31
Add function for unbiased random numbers

kurnevsky:
Later it will be used to choose random nodes for onion paths

### DIFF
--- a/src/toxcore/crypto_core.rs
+++ b/src/toxcore/crypto_core.rs
@@ -62,6 +62,19 @@ pub fn random_usize() -> usize {
     random_u64() as usize
 }
 
+/// Return unbiased random number from `[0, limit)` interval.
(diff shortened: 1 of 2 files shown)


kpp:
https://rust-random.github.io/rand/src/rand/distributions/uniform.rs.html#403

Would you please check it?

kurnevsky:
@kpp they do the same but with multiplication instead of modulo. Honestly, I don't know why. Perhaps it's more performant even though much more complex. In terms of correctness my version should be fine. Alternatively we can use `rand` crate instead of libsodium but I'd left it for the future.

kurnevsky:
@kpp right, it's because of performance: https://github.com/rust-random/rand/issues/570#issuecomment-408523182

kpp:
rebase

PR #314, opened 2019-02-11
Add ser/de of FILE_DATA packet

NamsooCho:


src/toxcore/messenger/packet/file_data.rs
```diff
@@ -0,0 +1,78 @@
...
+--------- | ------
+`1`       | `0x52`
+`1`       | `file number`
+`0..1371` | file data piece
+
+*/
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct FileData {
+    file_number: u8,
```
kpp: How about file_id?

kpp: Everything else looks good

NamsooCho: done

PR #316, opened 2019-02-13
Add ser/de of StatusMessage

NamsooCho:


src/toxcore/messenger/packet/status_message.rs
```diff
@@ -0,0 +1,88 @@
...
+Serialized form:
+
+Length    | Content
+--------- | ------
+`1`       | `0x31`
+`0..1007`  | UTF8 byte string
+
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StatusMessage(String);
```
kurnevsky: How about unifying with https://github.com/tox-rs/tox/pull/318/files#diff-8b40c8278e389268ea076fb2aaeb6d84R39 - you used short form of struct here, but full there.

PR #317, opened 2019-02-13
Add ser/de of UserStatus packet

NamsooCho:


src/toxcore/messenger/packet/user_status.rs
```diff
@@ -0,0 +1,81 @@
+/*! UserStatus struct.
+*/
+
+use nom::le_u8;
+
+use crate::toxcore::binary_io::*;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
+/// Status of me
```
kpp: of me? That's not a good doc for a struct named PeerStatus

src/toxcore/messenger/packet/user_status.rs
```diff
@@ -0,0 +1,81 @@
...
+    named!(from_bytes<PeerStatus>,
+        switch!(le_u8,
+            0 => value!(PeerStatus::Online) |
+            1 => value!(PeerStatus::Away) |
+            2 => value!(PeerStatus::Busy)
+        )
+    );
+}
+
+/** UserStatus is a struct that holds status of me.
```
kpp: of me?

src/toxcore/messenger/packet/user_status.rs
```diff
@@ -0,0 +1,81 @@
...
+Serialized form:
+
+Length    | Content
+--------- | ------
+`1`       | `0x32`
+`1`       | My status(0 = online, 1 = away, 2 = busy)
+
+*/
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct UserStatus {
```
kpp: Why don't you make UserStatus be enum?

```
pub enum UserStatus {
    /// Online
    Online = 0,
    /// Away
    Away,
    /// Online but I am busy
    Busy,
}
```

src/toxcore/messenger/packet/user_status.rs
```diff
@@ -0,0 +1,81 @@
+/*! UserStatus struct.
+*/
+
+use nom::le_u8;
+
+use crate::toxcore::binary_io::*;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
```
kurnevsky: What is it for?

src/toxcore/messenger/packet/user_status.rs
kurnevsky: What if we need to use status somewhere else? For example, state saving/loading.

src/toxcore/messenger/packet/user_status.rs
NamsooCho: It makes possible like this thing.
```
            gen_be_u8!(self.status as u8)
```

src/toxcore/messenger/packet/user_status.rs
kpp: Yes but business logic or state should not be described in /packets/ module. We'd better duplicate it somewhere else

kurnevsky: Why duplicate when we can move it to a common part?

kurnevsky: And it seems it's actually saved: https://github.com/TokTok/c-toxcore/blob/0aad180d1e1d7dae7cbf2868e139af0a4dc34fe7/toxcore/Messenger.c#L2812

kurnevsky: Also we will need to store friend's status in memory and return it to a client. So probably it can be stored somewhere in `toxcore/friends`. But anyway we can put it later to appropriate place, I think.

kpp: There should be different structs to store inside Saved_Friend and to send through the network.

NamsooCho: done

src/toxcore/messenger/packet/user_status.rs
kurnevsky: Why is it not possible without repr?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f61bd5f33844aecbf719174b567f1458

src/toxcore/messenger/packet/user_status.rs
```diff
@@ -0,0 +1,63 @@
...
+Length    | Content
+--------- | ------
+`1`       | `0x32`
+`1`       | User status(0 = online, 1 = away, 2 = busy)
+
+*/
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
+/// Status of user
+pub enum UserStatus {
```
kurnevsky: So what about this packet? Are we leaving it as a enum?

NamsooCho: I will revert it today.

NamsooCho: done

src/toxcore/messenger/packet/user_status.rs
```diff
@@ -0,0 +1,80 @@
+/*! UserStatus struct.
+*/
+
+use nom::le_u8;
+
+use crate::toxcore::binary_io::*;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+/// Status of user
```
kurnevsky: Not very important, but we put derives after docs everywhere. So I think it will look better here too.

NamsooCho: done

PR #323, opened 2019-02-19
Do not send the same packet to the same node twice

kurnevsky:
It's a bug introduced in `c-toxcore` after `assocs` refactoring. See the original code: https://github.com/irungentoo/toxcore/blob/bf69b54f64003d160d759068f4816b2d9b2e1e21/toxcore/DHT.c#L1740-L1763

Also see https://github.com/TokTok/c-toxcore/pull/1310

### DIFF
--- a/src/toxcore/dht/server/mod.rs
+++ b/src/toxcore/dht/server/mod.rs
@@ -545,7 +545,7 @@ impl Server {
             &self.pk,
             &payload
         ));
-        self.send_to_direct(node.saddr, ping_req)
+        self.send_to(node.saddr, ping_req)
     }
 
     /// Send `NodesRequest` packet to the node.
@@ -566,7 +566,7 @@ impl Server {
             &self.pk,
             &payload
         ));
-        Either::B(self.send_to_direct(node.saddr, nodes_req))
+        Either::B(self.send_to(node.saddr, nodes_req))
     }
 
     /// Send `NatPingRequest` packet to all friends and try to punch holes.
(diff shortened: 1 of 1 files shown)


src/toxcore/dht/server/mod.rs
```diff
@@ -635,9 +635,11 @@ impl Server {
     fn send_nat_ping_req_inner(&self, friend: &DhtFriend, nat_ping_req_packet: DhtRequest)
         -> impl Future<Item = (), Error = TimeoutError<mpsc::SendError<(Packet, SocketAddr)>>> + Send {
         let packet = Packet::DhtRequest(nat_ping_req_packet);
-        let futures = friend.close_nodes.nodes.iter().map(|node| {
-            self.send_to_node(node, &packet)
-        }).collect::<Vec<_>>();
+        let futures = friend.close_nodes.nodes
+            .iter()
+            .flat_map(|node| node.to_packed_node().into_iter())
```
kpp: How does it fix the issue?

kurnevsky: `to_packed_node` takes only one address from two and returns `Option`.

kpp: > takes only one address from two

based on IPv4/v6?

src/toxcore/dht/server/mod.rs
```diff
@@ -954,9 +943,9 @@ impl Server {
     fn handle_dht_req_for_others(&self, packet: DhtRequest)
         -> impl Future<Item = (), Error = HandlePacketError> + Send {
         let close_nodes = self.close_nodes.read();
-        if let Some(node) = close_nodes.get_node(&packet.rpk) { // search close_nodes to find target peer
+        if let Some(node) = close_nodes.get_node(&packet.rpk).and_then(|node| node.to_packed_node()) {
```
kpp: Better add one more combinator instead of `if let Some`:

```
.then(|maybe_node| {
   if let Some(node) = maybe_node {
   } else {
       Either::B(future::ok(()))

   }
}
```

src/toxcore/dht/server/mod.rs
kurnevsky: Based on last response.

kpp: OK, thanks

src/toxcore/dht/server/mod.rs
kurnevsky: Your code won't work - it has to be inner `if let` in this case which I don't like. I can move it to separate variable if you'd like.

kpp: > Your code won't work - it has to be inner `if let` in this case which I don't like.

Either you **don't** like it or code **won't** work. Please choose one option.

kurnevsky: It won't work just because `Option` doesn't have `then` method. So I had to make an assumption what did you mean an I supposed that it's two inner `if let`, but I can be wrong.

kpp: > just because `Option` doesn't have `then` method.

OK

PR #324, opened 2019-02-21
Add ser/de of FriendReq

NamsooCho:
The location of this file may be changed later

### DIFF
--- a/src/toxcore/onion/packet/friend_request.rs
+++ b/src/toxcore/onion/packet/friend_request.rs
@@ -1,0 +1,91 @@
+/*! FriendRequest packet
+*/
+
+use super::*;
+use std::str;
+use crate::toxcore::toxid::{NoSpam, NOSPAMBYTES};
+use crate::toxcore::friend_connection::packet::MAX_ONION_CLIENT_DATA_SIZE;
+
+const MAX_FRIEND_REQUEST_DATA_SIZE: usize = MAX_ONION_CLIENT_DATA_SIZE - (1 + NOSPAMBYTES);
+
+/** Friend request that can be enclosed in onion data packet and sent through onion
+path.
+
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct FriendRequest {
+    nospam: NoSpam,
+    msg: String,
+}
+
+impl ToBytes for FriendRequest {
+    fn to_bytes<'a>(&self, buf: (&'a mut [u8], usize)) -> Result<(&'a mut [u8], usize), GenError> {
+        do_gen!(buf,
+            gen_be_u8!(0x20) >>
+            gen_slice!(self.nospam.0) >>
+            gen_cond!(self.msg.len() > MAX_FRIEND_REQUEST_DATA_SIZE, |buf| gen_error(buf, 0)) >>
+            gen_slice!(self.msg.as_bytes())
+        )
+    }
+}
+
+impl FromBytes for FriendRequest {
+    named!(from_bytes<FriendRequest>, do_parse!(
+        tag!(&[0x20][..]) >>
+        nospam: call!(NoSpam::from_bytes) >>
+        msg: map_res!(verify!(rest, |msg: &[u8]| msg.len() <= MAX_FRIEND_REQUEST_DATA_SIZE), str::from_utf8) >>
+        (FriendRequest {
+            nospam,
+            msg: msg.to_string(),
+        })
+    ));
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    encode_decode_test!(
+        friend_req_encode_decode,
+        FriendRequest {
+            nospam: NoSpam::random(),
+            msg: "1234".to_owned(),
+        }
+    );
+
+    // Test for encoding error of from_bytes.
+    #[test]
+    fn friend_req_from_bytes_encoding_error() {
+        let err_string = vec![0, 159, 146, 150]; // not UTF8 bytes.
+        let nospam = NoSpam::random();
+
+        let mut friend_req = vec![0x20];
+        friend_req.extend_from_slice(&nospam.0);
+        friend_req.extend(err_string);
+        assert!(FriendRequest::from_bytes(&friend_req).is_err());
+    }
+
+    // Test for overflow of from_bytes.
+    #[test]
+    fn friend_req_from_bytes_overflow() {
+        let large_string = vec![32; MAX_FRIEND_REQUEST_DATA_SIZE + 1];
+        let nospam = NoSpam::random();
+
+        let mut friend_req = vec![0x20];
+        friend_req.extend_from_slice(&nospam.0);
+        friend_req.extend(large_string);
+        assert!(FriendRequest::from_bytes(&friend_req).is_err());
+    }
+
+    // Test for overflow of to_bytes.
+    #[test]
+    fn friend_req_to_bytes_overflow() {
+        let large_string = String::from_utf8(vec![32u8; MAX_FRIEND_REQUEST_DATA_SIZE + 1]).unwrap();
+        let friend_req = FriendRequest {
+            nospam: NoSpam::random(),
+            msg: large_string
+        };
+        let mut buf = [0; MAX_FRIEND_REQUEST_DATA_SIZE + 1]; // `1` is for packet_id.
+        assert!(friend_req.to_bytes((&mut buf, 0)).is_err());
+    }
+}
--- a/src/toxcore/onion/packet/friend_request.rs
+++ b/src/toxcore/onion/packet/friend_request.rs
@@ -4,12 +4,18 @@
 use super::*;
 use std::str;
 use crate::toxcore::toxid::{NoSpam, NOSPAMBYTES};
-use crate::toxcore::friend_connection::packet::MAX_ONION_CLIENT_DATA_SIZE;
+use crate::toxcore::friend_connection::packet::*;
 
-const MAX_FRIEND_REQUEST_DATA_SIZE: usize = MAX_ONION_CLIENT_DATA_SIZE - (1 + NOSPAMBYTES);
+const MAX_FRIEND_REQUEST_MSG_SIZE: usize = MAX_ONION_CLIENT_DATA_SIZE - (1 + NOSPAMBYTES);
 
 /** Friend request that can be enclosed in onion data packet and sent through onion
 path.
+
+Length    | Content
+--------- | -------------------------
+`1`       | `0x20`
+`4`       | NoSpam
+`1..991`  | Message
 
 */
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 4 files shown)


kpp:
How about this https://github.com/tox-rs/tox/blob/onion_client/src/toxcore/onion/packet/onion_data_response.rs#L69 ?

kpp:
Sorry for closing the PR. Pushed wrong button =)

NamsooCho:
At least,
https://github.com/tox-rs/tox/blob/2e7ac6c31845bf9dc85e0aa59bd4a7443ab926a0/src/toxcore/onion/packet/onion_data_response.rs#L69

payload parameter for `new` is a successful result of `get_payload`, so it would not call `new` with payload if the result of `get_payload` is Err. When we call `get_payload`, we must handle Err.

kurnevsky:
Can `FriendReq` be bigger than `OnionDataResponse`?

src/toxcore/onion/packet/friend_req.rs
```diff
@@ -0,0 +1,89 @@
...
+use std::str;
+use crate::toxcore::toxid::NoSpam;
+use crate::toxcore::friend_connection::packet::MAX_ONION_CLIENT_DATA_SIZE;
+
+/** Friend request that can be enclosed in onion data packet and sent through onion
+path.
+
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct FriendReq {
```
kurnevsky: I'd call it `FriendRequest`, we use full words for other packets.

NamsooCho: done

src/toxcore/onion/packet/friend_request.rs
```diff
@@ -0,0 +1,91 @@
+/*! FriendRequest packet
+*/
+
+use super::*;
+use std::str;
+use crate::toxcore::toxid::{NoSpam, NOSPAMBYTES};
+use crate::toxcore::friend_connection::packet::MAX_ONION_CLIENT_DATA_SIZE;
+
+const MAX_FRIEND_REQUEST_DATA_SIZE: usize = MAX_ONION_CLIENT_DATA_SIZE - (1 + NOSPAMBYTES);
```
kpp: MAX_FRIEND_REQUEST_**MSG**_SIZE?

src/toxcore/onion/packet/friend_request.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    // Test for overflow of to_bytes.
+    #[test]
+    fn friend_req_to_bytes_overflow() {
+        let large_string = String::from_utf8(vec![32u8; MAX_FRIEND_REQUEST_DATA_SIZE + 1]).unwrap();
+        let friend_req = FriendRequest {
+            nospam: NoSpam::random(),
+            msg: large_string
+        };
+        let mut buf = [0; MAX_FRIEND_REQUEST_DATA_SIZE + 1]; // `1` is for packet_id.
```
kpp: replace MAX_FRIEND_REQUEST_DATA_SIZE + 1 -> MAX_ONION_CLIENT_DATA_SIZE

src/toxcore/onion/packet/onion_data_response.rs
```diff
@@ -66,7 +66,7 @@ impl OnionDataResponse {
     /// Create `OnionDataResponse` from `OnionDataResponsePayload` encrypting it
     /// with `shared_key` and `nonce`
     pub fn new(shared_secret: &PrecomputedKey, temporary_pk: PublicKey, nonce: Nonce, payload: &OnionDataResponsePayload) -> OnionDataResponse {
-        let mut buf = [0; 293]; // TODO: check FriendRequest size
+        let mut buf = [0; 293];
```
kpp: we should not use 293 magic constant

src/toxcore/onion/packet/onion_data_response.rs
```diff
@@ -153,7 +153,7 @@ impl OnionDataResponsePayload {
     /// Create `OnionDataResponsePayload` from `OnionDataResponseInnerPayload`
     /// encrypting it with `shared_key` and `nonce`
     pub fn new(shared_secret: &PrecomputedKey, real_pk: PublicKey, nonce: &Nonce, payload: &OnionDataResponseInnerPayload) -> OnionDataResponsePayload {
-        let mut buf = [0; 245]; // TODO: check FriendRequest size
+        let mut buf = [0; 245];
```
kpp: we should not use 245 magic constant

src/toxcore/onion/packet/onion_data_response.rs
```diff
@@ -199,19 +199,22 @@ impl OnionDataResponsePayload {
 pub enum OnionDataResponseInnerPayload {
     /// [`DhtPkAnnouncePayload`](../../dht/packet/struct.DhtPkAnnouncePayload.html) structure.
     DhtPkAnnounce(DhtPkAnnouncePayload),
-    // TODO: FriendRequest
+    /// [`FriendRequest`](../../dht/packet/struct.FriendRequest.html) structure.
+    FriendRequest(FriendRequest),
```
kpp: Should not we add tests for OnionDataResponseInnerPayload with FriendRequest, at least a successful condition? (example of unsuccessful for DhtPkAnnounce in onion_data_response_payload_encrypt_decrypt_invalid_key)

NamsooCho: done

src/toxcore/onion/packet/friend_request.rs
```diff
@@ -18,12 +18,22 @@ pub struct FriendRequest {
...
+    }
+}
+
 impl ToBytes for FriendRequest {
     fn to_bytes<'a>(&self, buf: (&'a mut [u8], usize)) -> Result<(&'a mut [u8], usize), GenError> {
         do_gen!(buf,
             gen_be_u8!(0x20) >>
             gen_slice!(self.nospam.0) >>
-            gen_cond!(self.msg.len() > MAX_FRIEND_REQUEST_DATA_SIZE, |buf| gen_error(buf, 0)) >>
+            gen_cond!(self.msg.len() > MAX_FRIEND_REQUEST_MSG_SIZE && self.msg.len() > 0, |buf| gen_error(buf, 0)) >>
```
kpp: > `self.msg.len() > MAX_FRIEND_REQUEST_MSG_SIZE && self.msg.len() > 0`

Please build https://en.wikipedia.org/wiki/Karnaugh_map for this expression.

src/toxcore/onion/packet/friend_request.rs
```diff
@@ -80,12 +90,12 @@ mod tests {
...
     #[test]
     fn friend_req_to_bytes_overflow() {
-        let large_string = String::from_utf8(vec![32u8; MAX_FRIEND_REQUEST_DATA_SIZE + 1]).unwrap();
+        let large_string = String::from_utf8(vec![32u8; MAX_FRIEND_REQUEST_MSG_SIZE + 1]).unwrap();
         let friend_req = FriendRequest {
             nospam: NoSpam::random(),
             msg: large_string
         };
-        let mut buf = [0; MAX_FRIEND_REQUEST_DATA_SIZE + 1]; // `1` is for packet_id.
+        let mut buf = [0; MAX_ONION_CLIENT_DATA_SIZE + 1]; // `1` is to provide enough space for success of serializing
```
kpp: No need to make `+ 1`, because we will for sure serialize in into buf with size = MAX_ONION_CLIENT_DATA_SIZE

NamsooCho: https://github.com/tox-rs/tox/blob/ab27c8d11c4f02944f7ff5d30ce986ad4c56be49/src/toxcore/onion/packet/friend_request.rs#L93

My intention is to provide even if ser was success.

src/toxcore/onion/packet/onion_data_response.rs
```diff
@@ -66,7 +70,7 @@ impl OnionDataResponse {
     /// Create `OnionDataResponse` from `OnionDataResponsePayload` encrypting it
     /// with `shared_key` and `nonce`
     pub fn new(shared_secret: &PrecomputedKey, temporary_pk: PublicKey, nonce: Nonce, payload: &OnionDataResponsePayload) -> OnionDataResponse {
-        let mut buf = [0; 293];
```
kpp: 293 - 245 = 48

MAX_ONION_RESPONSE_PAYLOAD_SIZE = 293 (correct?)
MAX_ONION_RESPONSE_PAYLOAD_SIZE = MAX_ONION_CLIENT_DATA_SIZE + PUBLICKEYBYTES

MAX_ONION_CLIENT_DATA_SIZE = 245 (correct?)

so PUBLICKEYBYTES = MAX_ONION_RESPONSE_PAYLOAD_SIZE - MAX_ONION_CLIENT_DATA_SIZE = 48,

but 

```
#define crypto_box_curve25519xsalsa20poly1305_PUBLICKEYBYTES 32U
```

Where is the truth?

NamsooCho: 293 and 245 are for DhtPkAnnounce only.
These numbers is correct if we use only DhtPkAnnounce.
Now, we added FriendRequest, so this array must be entended.

src/toxcore/onion/packet/friend_request.rs
kpp: Ah I see. Good job.

src/toxcore/onion/packet/friend_request.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        let mut friend_req = vec![0x20];
+        friend_req.extend_from_slice(&nospam.0);
+        friend_req.extend(err_string);
+        assert!(FriendRequest::from_bytes(&friend_req).is_err());
+    }
+
+    // Test for overflow of from_bytes.
+    #[test]
+    fn friend_req_from_bytes_overflow() {
```
kpp: test for underflow is missing (when msg size = 0)

kurnevsky:
Something is not right. `ONION_CLIENT_MAX_DATA_SIZE` in `c-toxcore` has value `1021` while our `MAX_ONION_CLIENT_DATA_SIZE` has value `996`.

NamsooCho:
const difference is fixed.

kurnevsky:
Something is still not right. MAX_ONION_RESPONSE_PAYLOAD_SIZE-MAX_ONION_CLIENT_DATA_SIZE=1053-1021=32, but the difference between `OnionDataResponse` buffer length and `OnionDataResponsePayload` buffer length should be 48.  `OnionDataResponsePayload` buffer includes only serialized `OnionDataResponseInnerPayload` while `OnionDataResponse` buffer includes _encrypted_ `OnionDataResponseInnerPayload` and long term `PublicKey` which gives key+mac=32+16=48.

kurnevsky:
If I got it right `MACBYTES` are missing here:
```rust
const MAX_ONION_RESPONSE_PAYLOAD_SIZE: usize = MAX_ONION_CLIENT_DATA_SIZE + PUBLICKEYBYTES;
```

NamsooCho:
Thanks! fixing is done.

PR #325, opened 2019-02-22
Good to ping

NamsooCho:
Add function to check if we need to send ping.
function name is `is_redundant_ping`

### DIFF
--- a/src/toxcore/onion/onion_client.rs
+++ b/src/toxcore/onion/onion_client.rs
@@ -64,6 +64,8 @@ const ONION_DHTPK_SEND_INTERVAL: u64 = 30;
 /// via DHT request.
 const DHT_DHTPK_SEND_INTERVAL: u64 = 20;
 
+const MIN_NODE_PING_TIME: u64 = 10;
+
 #[derive(Clone, Debug)]
 struct OnionFriend {
     /// Friend's long term `PublicKey`.
@@ -292,6 +294,22 @@ impl OnionClient {
         }
     }
 
+    fn is_redundant_ping(pk: PublicKey, search_pk: PublicKey, request_queue: &RequestQueue<AnnounceRequestData>) -> bool {
+        let check_pks = |data: &AnnounceRequestData| -> bool {
+            if let Some(friend_pk) = data.friend_pk {
+                data.pk == pk && search_pk == friend_pk
+            } else {
+                data.pk == pk
+            }
+        };
+        let request = request_queue.find(check_pks);
+        if let Some((ping_time, _request_data)) = request {
+            clock_elapsed(*ping_time) < Duration::from_secs(MIN_NODE_PING_TIME)
+        } else {
+            false
+        }
+    }
+
     /// Handle `OnionAnnounceResponse` packet.
     pub fn handle_announce_response(&self, packet: &OnionAnnounceResponse, _addr: SocketAddr) -> Box<Future<Item = (), Error = Error> + Send> {
         let state = &mut *self.state.write();
@@ -358,9 +376,12 @@ impl OnionClient {
         let mut futures = Vec::with_capacity(payload.nodes.len());
 
         for node in &payload.nodes {
-            // TODO: reduce pings frequency? (see good_to_ping in c-toxcore)
-
             if !nodes_list.can_add(&announce_packet_data.search_pk, &node, /* evict */ true) {
+                continue;
+            }
+
+            // To prevent to send redundant ping packet.
+            if OnionClient::is_redundant_ping(node.pk, announce_packet_data.search_pk, &state.announce_requests) {
                 continue;
             }
 
--- a/src/toxcore/onion/onion_client.rs
+++ b/src/toxcore/onion/onion_client.rs
@@ -294,20 +294,18 @@ impl OnionClient {
         }
     }
 
-    fn is_redundant_ping(pk: PublicKey, search_pk: PublicKey, request_queue: &RequestQueue<AnnounceRequestData>) -> bool {
+    fn is_redundant_ping(&self, pk: PublicKey, search_pk: PublicKey, request_queue: &RequestQueue<AnnounceRequestData>) -> bool {
         let check_pks = |data: &AnnounceRequestData| -> bool {
-            if let Some(friend_pk) = data.friend_pk {
-                data.pk == pk && search_pk == friend_pk
+            let request_search_pk = if let Some(friend_pk) = data.friend_pk {
+                friend_pk
             } else {
-                data.pk == pk
-            }
+                self.dht.pk
+            };
+            data.pk == pk && search_pk == request_search_pk
         };
-        let request = request_queue.find(check_pks);
-        if let Some((ping_time, _request_data)) = request {
-            clock_elapsed(*ping_time) < Duration::from_secs(MIN_NODE_PING_TIME)
-        } else {
-            false
-        }
+        request_queue.get_values()
+            .any(|(ping_time, request_data)| check_pks(request_data) &&
+                clock_elapsed(ping_time) < Duration::from_secs(MIN_NODE_PING_TIME))
     }
 
     /// Handle `OnionAnnounceResponse` packet.
@@ -381,7 +379,7 @@ impl OnionClient {
             }
 
             // To prevent to send redundant ping packet.
-            if OnionClient::is_redundant_ping(node.pk, announce_packet_data.search_pk, &state.announce_requests) {
+            if self.is_redundant_ping(node.pk, announce_packet_data.search_pk, &state.announce_requests) {
                 continue;
             }
 
--- a/src/toxcore/onion/request_queue.rs
+++ b/src/toxcore/onion/request_queue.rs
@@ -51,16 +51,11 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
         );
     }
 
-    /// Returns the Instant of a HashMap according to `f`.
-    pub fn find<F>(&self, f: F) -> Option<(&Instant, &T)>
-        where F: Fn(&T) -> bool {
-        self.ping_map.iter()
-            .find_map(|(_ping_id, (ping_time, data))| {
-                if f(data) {
-                    Some((ping_time, data))
-                } else {
-                    None
-                }
-            })
+    /// Returns the Values of ping_map which is not timedout.
+    pub fn get_values(&self) -> impl Iterator<Item=(Instant, &T)> {
+        self.ping_map
+            .values()
+            .filter(move |(time, _request_data)| clock_elapsed(*time) <= self.timeout)
+            .map(|(time, data)| (time.clone(), data))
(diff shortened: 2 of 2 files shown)


src/toxcore/onion/request_queue.rs
```diff
@@ -6,7 +6,7 @@ use crate::toxcore::time::*;
 
 #[derive(Clone, Debug)]
 pub struct RequestQueue<T> {
-    ping_map: HashMap<u64, (Instant, T)>,
+    pub ping_map: HashMap<u64, (Instant, T)>,
```
kurnevsky: Add method that will return optional timestamp rather than making it public.

src/toxcore/onion/onion_client.rs
```diff
@@ -292,6 +292,25 @@ impl OnionClient {
...
+                if request_data.pk == pk {
+                    Some(time)
+                } else {
+                    None
+                }
+            })
+            .filter(|time| clock_elapsed(**time) < Duration::from_secs(ONION_NODE_PING_INTERVAL))
+            .collect::<Vec<_>>();
+
+        if last_ping_time.is_empty() {
```
kurnevsky: !last_ping_time.is_empty()

kurnevsky: Change it to `map_or` after method adding.

src/toxcore/onion/request_queue.rs
```diff
@@ -50,4 +50,28 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
...
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let len = ping_times.len();
+        if len == 0 {
+            None
+        } else if len > 1 {
+            warn!("There are multiple same pings to a peer");
+            None
```
kurnevsky: Do we need this warn? Why don't just use `find_map`? Also returning `None` here will lead to more duplications.

src/toxcore/onion/request_queue.rs
```diff
@@ -50,4 +50,28 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
             clock_elapsed(time) <= timeout
         );
     }
+
+    /// Returns the Instant of a PublicKey, used to check if the PK is timedout
+    pub fn get_ping_time<F>(&self, f: F) -> Option<&Instant>
```
kurnevsky: Maybe make it more generic, return `&(Instant, T)` and call it just `find`?

src/toxcore/onion/onion_client.rs
```diff
@@ -292,6 +292,22 @@ impl OnionClient {
...
+        let check_pks = |data: &AnnounceRequestData| -> bool {
+            if let Some(friend_pk) = data.friend_pk {
+                data.pk == pk && search_pk == friend_pk
+            } else {
+                data.pk == pk
+            }
+        };
+        let ping_time = request_queue.get_ping_time(check_pks);
+        if let Some(ping_time) = ping_time {
+            clock_elapsed(*ping_time) < Duration::from_secs(ONION_NODE_PING_INTERVAL)
```
kurnevsky: It should be `MIN_NODE_PING_TIME = 10`.

NamsooCho: done

src/toxcore/onion/request_queue.rs
```diff
@@ -50,4 +50,17 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
             clock_elapsed(time) <= timeout
         );
     }
+
+    /// Returns the Instant of a HashMap according to `f`.
+    pub fn find<F>(&self, f: F) -> Option<(&Instant, &T)>
+        where F: Fn(&T) -> bool {
+        self.ping_map.iter()
+            .find_map(|(_ping_id, (ping_time, data))| {
```
kurnevsky: Use `ping_map.values` since you don't need `ping_id` here.

src/toxcore/onion/onion_client.rs
```diff
@@ -292,6 +294,22 @@ impl OnionClient {
         }
     }
 
+    fn is_redundant_ping(pk: PublicKey, search_pk: PublicKey, request_queue: &RequestQueue<AnnounceRequestData>) -> bool {
+        let check_pks = |data: &AnnounceRequestData| -> bool {
+            if let Some(friend_pk) = data.friend_pk {
+                data.pk == pk && search_pk == friend_pk
+            } else {
+                data.pk == pk
```
kurnevsky: Here you should check that `search_pk` equals our `dht_pk`, otherwise you can find previous announce request for current search request. Then it can be rewritten to:
```rust
let request_search_pk = if Some(friend_pk) = data.friend_pk {
    friend_pk
} else {
    dht_pk
};
data.pk == pk && search_pk == request_search_pk
```
I think it looks better since here you just check that npde's pk and search pk from new request equals to same pks from old request.

NamsooCho: done

src/toxcore/onion/onion_client.rs
```diff
@@ -358,12 +377,15 @@ impl OnionClient {
...
 
         for node in &payload.nodes {
-            // TODO: reduce pings frequency? (see good_to_ping in c-toxcore)
-
             if !nodes_list.can_add(&announce_packet_data.search_pk, &node, /* evict */ true) {
                 continue;
             }
 
+            // To prevent to send redundant ping packet.
+            if self.is_redundant_ping(node.pk, announce_packet_data.search_pk, &state.announce_requests) {
```
kpp: Why did you add self as the first arg? It was not here before. Try to call clippy  on the project

NamsooCho: https://github.com/tox-rs/tox/blob/00a77960257b865aa29fc4e662817d9021498a68/src/toxcore/onion/onion_client.rs#L302

Now self is needed.

NamsooCho:
>How about .filter instead of find_map
done

src/toxcore/onion/request_queue.rs
```diff
@@ -50,4 +50,14 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
...
     }
+
+    /// Returns the Instant of a HashMap according to `f`.
+    pub fn find<F>(&self, f: F) -> Option<&(Instant, T)>
+        where F: Fn(&T) -> bool {
+        self.ping_map.values()
+            .filter(|(_ping_time, data)| {
+                f(data)
+            })
+            .next()
```
kurnevsky: Then it can be just `find` :)

NamsooCho: done

src/toxcore/onion/onion_client.rs
kpp: OK

src/toxcore/onion/request_queue.rs
```diff
@@ -50,4 +50,11 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
             clock_elapsed(time) <= timeout
         );
     }
+
+    /// Returns the Instant of a HashMap according to `f`.
+    pub fn find<F>(&self, f: F) -> Option<&(Instant, T)>
```
kpp: find_by_value

src/toxcore/onion/onion_client.rs
```diff
@@ -292,6 +294,25 @@ impl OnionClient {
...
+        let check_pks = |data: &AnnounceRequestData| -> bool {
+            let request_search_pk = if let Some(friend_pk) = data.friend_pk {
+                friend_pk
+            } else {
+                self.dht.pk
+            };
+            data.pk == pk && search_pk == request_search_pk
+        };
+        let request = request_queue.get_values()
+            .filter(|(ping_time, _request_data)| clock_elapsed(*ping_time) <= request_queue.get_timeout())
```
kurnevsky: Wrong place to filter timed out records. It doesn't make sense just because next you check time anyway. It rather should be [here](https://github.com/tox-rs/tox/pull/325/files#diff-6e41041a9b7f108d886020985a2290c2R57) to make this function consistent regardless of `clear_timed_out` execution.

kpp: Why?

src/toxcore/onion/request_queue.rs
```diff
@@ -50,4 +50,16 @@ impl<T> RequestQueue<T> { // TODO: unify with DHT?
             clock_elapsed(time) <= timeout
         );
     }
+
+    /// Returns the Values of ping_map.
+    pub fn get_values(&self) -> impl Iterator<Item=(Instant, &T)> {
+        self.ping_map
+            .values()
```
kpp: Namsoo, add `.filter(|(ping_time, _request_data)| clock_elapsed(*ping_time) <= request_queue.get_timeout())` here, and add mark to doc comment that we return fresh values

NamsooCho: done

src/toxcore/onion/onion_client.rs
```diff
@@ -292,6 +294,24 @@ impl OnionClient {
...
+            let request_search_pk = if let Some(friend_pk) = data.friend_pk {
+                friend_pk
+            } else {
+                self.dht.pk
+            };
+            data.pk == pk && search_pk == request_search_pk
+        };
+        let request = request_queue.get_values()
+            .find(|(_ping_time, request_data)| check_pks(request_data));
+        if let Some((ping_time, _request_data)) = request {
```
kpp: Now the check for timeout is redundant here, check with is_some

NamsooCho: As I know, the timeout of RequestQueue is the timeout of Queue itself and this check is for preventing redundant ping.

NamsooCho: RequestQueue timeout is ANNOUNCE_TIMEOUT which is 300 seconds and cheking timeout is 10 seconds.

src/toxcore/onion/onion_client.rs
```diff
@@ -292,6 +294,24 @@ impl OnionClient {
...
+        let check_pks = |data: &AnnounceRequestData| -> bool {
+            let request_search_pk = if let Some(friend_pk) = data.friend_pk {
+                friend_pk
+            } else {
+                self.dht.pk
+            };
+            data.pk == pk && search_pk == request_search_pk
+        };
+        let request = request_queue.get_values()
+            .find(|(_ping_time, request_data)| check_pks(request_data));
```
kurnevsky: After yesterday's discussion with @kpp I realized that this is not a correct check. We have different timeouts for allowing new requests and for expiration of old requests. So it's possible that we have different entries in the queue for the same request. Which means instead of finding the first one we should check all matches, i.e. instead of `find` we should use `any` or `all`. Otherwise we can send a new request even though we sent it recently.

NamsooCho: done

PR #330, opened 2019-03-02
Fix main loop interval for net crypto

kurnevsky:


src/toxcore/net_crypto/mod.rs
```diff
@@ -741,7 +741,7 @@ impl NetCrypto {
     /// Run `net_crypto` periodical tasks. Result future will never be completed
     /// successfully.
     pub fn run(self) -> impl Future<Item = (), Error = RunError> + Send {
-        let interval = Duration::from_secs(PACKET_COUNTER_AVERAGE_INTERVAL);
+        let interval = Duration::from_millis(PACKET_COUNTER_AVERAGE_INTERVAL);
```
kpp: PACKET_COUNTER_AVERAGE_INTERVAL -> PACKET_COUNTER_AVERAGE_INTERVAL_MS?

kurnevsky: I think we should use `Duration` as const type instead of just `u64`, and add `_MS` postfix where we need millis. It should be done everywhere, not just for `PACKET_COUNTER_AVERAGE_INTERVAL`. So should I change it in this PR?

kpp: Yes please!

PR #333, opened 2019-03-04
Make get_closest function public and parameterize it with count

kurnevsky:


kpp:
Why?

kurnevsky:
https://github.com/tox-rs/tox/blob/8950d53aece81e4bc2fbb81caedea43b26740712/src/toxcore/onion/onion_client.rs#L582

PR #334, opened 2019-03-04
Add method for removing a friend

kurnevsky:


src/toxcore/dht/server/mod.rs
```diff
@@ -288,6 +288,12 @@ impl Server {
         friends.insert(friend_pk, friend);
     }
 
+    /// Remove a friend from the DHT friends list to stop looking for it's IP
+    /// address.
+    pub fn remove_friend(&self, friend_pk: PublicKey) {
```
kpp: Do we need to return any result to the user?

kurnevsky: Not sure. The final state matters, and I think we won't delete nonexistent friends. So only if we need to assert it.

PR #337, opened 2019-03-05
Reduce the number of tokio deps

kpp:


Cargo.toml
```diff
@@ -34,7 +34,6 @@ byteorder = "1"
 futures = "0.1"
 log = "0.4"
 sodiumoxide = "0.2"
-tokio = "0.1"
 tokio-codec = "0.1"
```
kurnevsky: Do we need tokio-codec here?

kpp: Fixed

PR #338, opened 2019-03-06
Adjust RequestQueue so that it can be used for onion client

kurnevsky:


src/toxcore/dht/request_queue.rs
```diff
@@ -1,71 +1,85 @@
...
+            return None;
         }
 
-        match self.ping_map.remove(&(pk, ping_id)) {
-            Some(time) if clock_elapsed(time) <= self.timeout => true,
-            _ => false,
+        if let Entry::Occupied(entry) = self.ping_map.entry(ping_id) {
+            let &(time, ref data) = entry.get();
+            if clock_elapsed(time) <= self.timeout && cond(data) {
+                Some((entry.remove_entry().1).1)
```
kpp: Rewrite `(entry.remove_entry().1).1` to temp variables

src/toxcore/dht/request_queue.rs
```diff
@@ -1,71 +1,85 @@
...
         if ping_id == 0 {
-            return false
+            return None;
         }
 
-        match self.ping_map.remove(&(pk, ping_id)) {
-            Some(time) if clock_elapsed(time) <= self.timeout => true,
-            _ => false,
+        if let Entry::Occupied(entry) = self.ping_map.entry(ping_id) {
+            let &(time, ref data) = entry.get();
```
kpp: Can you try to write it without `let &()`?

src/toxcore/dht/request_queue.rs
```diff
@@ -1,71 +1,85 @@
...
+        } else {
+            None
         }
     }
 
     /// Remove timed out request IDs.
     pub fn clear_timed_out(&mut self) {
         let timeout = self.timeout;
-        self.ping_map.retain(|&_, &mut time|
+        self.ping_map.retain(|&_, &mut (time, _)|
```
kpp: `&mut`?

src/toxcore/dht/request_queue.rs
```diff
@@ -79,80 +93,69 @@ mod tests {
...
-        let ping_id = queue.new_ping_id(pk);
+        let ping_id = queue.new_ping_id(7);
 
-        assert!(queue.ping_map.contains_key(&(pk, ping_id)));
+        assert_eq!(queue.ping_map[&ping_id].1, 7);
     }
 
     #[test]
     fn check_ping_id() {
-        crypto_init().unwrap();
```
kpp: gen_ping_id calls to random_u32 which requires crypto_init()?

src/toxcore/dht/request_queue.rs
kurnevsky: Why? And how?

src/toxcore/dht/request_queue.rs
kurnevsky: Yes, it's the signature of `retain` method - it takes such closure.

src/toxcore/dht/request_queue.rs
kurnevsky: You are right, fixed.

src/toxcore/dht/request_queue.rs
kurnevsky: done

src/toxcore/dht/request_queue.rs
kpp:
```
            let (time, data) = entry.get();
            if clock_elapsed(*time) <= self.timeout && cond(data) {
```

src/toxcore/dht/request_queue.rs
```diff
@@ -1,71 +1,86 @@
...
             clock_elapsed(time) <= timeout
         );
     }
+
+    /// Get not timed out requests stored in this `RequestQueue`.
+    pub fn get_values(&self) -> impl Iterator<Item = (Instant, &T)> {
+        self.ping_map
+            .values()
+            .filter(move |(time, _)| clock_elapsed(*time) <= self.timeout)
+            .map(|&(time, ref data)| (time, data))
```
kpp: .map(|(time, data)| (*time, data))

kurnevsky: Why?

src/toxcore/dht/request_queue.rs
kurnevsky: Why?

kpp: Because I think it is easier to read than your version.

src/toxcore/dht/request_queue.rs
kpp: Because I think it is easier to read than your version.

PR #339, opened 2019-03-06
More generic KbucketNode

kurnevsky:


src/toxcore/dht/dht_node.rs
```diff
@@ -202,6 +202,7 @@ impl HasPK for DhtNode {
 
 impl KbucketNode for DhtNode {
     type NewNode = PackedNode;
+    type CheckNode = PackedNode;
```
kpp: Why?

kurnevsky: Usually to check if a node can be added to a kbucket we need less data that we actually add it. So in onion client I check PackedNode because I don't want to construct the whole OnionNode - I just don't have all the data at that point: https://github.com/tox-rs/tox/blob/8950d53aece81e4bc2fbb81caedea43b26740712/src/toxcore/onion/onion_client.rs#L377

PR #343, opened 2019-03-13
Add function to get random relays we are connected to

kurnevsky:
We should send them to our friends.

### DIFF
--- a/src/toxcore/tcp/client/connections.rs
+++ b/src/toxcore/tcp/client/connections.rs
@@ -25,6 +25,7 @@ use futures::sync::mpsc;
 use tokio::timer::Interval;
 
 use crate::toxcore::crypto_core::*;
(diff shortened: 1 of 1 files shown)


src/toxcore/tcp/client/connections.rs
```diff
@@ -279,6 +280,25 @@ impl Connections {
...
+            .values()
+            .filter(|client| client.is_connected())
+            .map(|client| PackedNode::new(client.addr, &client.pk))
+            .collect::<Vec<_>>();
+
+        if relays.is_empty() {
+            return Vec::new();
+        }
+
+        // TODO: shuffle relays instead
```
kpp: Why todo?

kurnevsky: Because with libsodium we don't have a function for random shuffle. It can be done later with `rand` crate.

PR #346, opened 2019-03-16
Add function to set friend's IP address

kurnevsky:
Also added optional IP address to `add_connection` as it might be known at the moment we add new connection.

### DIFF
--- a/src/toxcore/net_crypto/mod.rs
+++ b/src/toxcore/net_crypto/mod.rs
@@ -194,7 +194,7 @@ impl NetCrypto {
     }
 
     /// Add connection to a friend when its DHT `PublicKey` is known.
-    pub fn add_connection(&self, peer_real_pk: PublicKey, peer_dht_pk: PublicKey) {
+    pub fn add_connection(&self, peer_real_pk: PublicKey, peer_dht_pk: PublicKey, saddr: Option<SocketAddr>) {
         let mut connections = self.connections.write();
 
         if connections.contains_key(&peer_real_pk) {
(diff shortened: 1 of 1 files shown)


src/toxcore/net_crypto/mod.rs
```diff
@@ -194,23 +194,48 @@ impl NetCrypto {
...
         let dht_precomputed_key = precompute(&peer_dht_pk, &self.dht_sk);
-        let connection = CryptoConnection::new(
+        let mut connection = CryptoConnection::new(
             &dht_precomputed_key,
             self.dht_pk,
             self.real_pk,
             peer_real_pk,
             peer_dht_pk
         );
+        connection.udp_addr = saddr;
```
kpp: Is there any code where you don't set `connection.udp_addr` right after you have created `CryptoConnection::new`?

src/toxcore/net_crypto/mod.rs
```diff
@@ -194,23 +194,48 @@ impl NetCrypto {
...
         let connection = Arc::new(RwLock::new(connection));
         connections.insert(peer_real_pk, connection);
+        if let Some(saddr) = saddr {
+            self.keys_by_addr.write().insert((saddr.ip(), saddr.port()), peer_real_pk);
```
kpp: How about `self.set_friend_udp_addr`?

src/toxcore/net_crypto/mod.rs
kurnevsky: Yes. It might not be present in Ktree after we found a DHT key. In this case we will find it later. That's why `saddr` is `Option`.

PR #347, opened 2019-03-19
Store both IPv4 and IPv6 addresses in the crypto connection

kurnevsky:


src/toxcore/net_crypto/crypto_connection.rs
```diff
@@ -234,6 +236,35 @@ impl RecvPacket {
     }
 }
 
+/// UDP address of a connection with the time when last UDP packet was received
+#[derive(Clone, Debug, PartialEq)]
+pub struct ConnectionAddr<T: Into<SocketAddr> + Copy> {
+    /// Address to send UDP packets directly to the peer
+    pub addr: T,
+    /// Time when last UDP packet was received
+    pub received_time: Instant,
```
kpp: last_received_time?

src/toxcore/net_crypto/crypto_connection.rs
```diff
@@ -433,9 +464,60 @@ impl CryptoConnection {
         }
     }
 
-    /// Set time when last UDP packet was received to now
-    pub fn update_udp_received_time(&mut self) {
-        self.udp_received_time = Some(clock_now())
+    /// Set UPD address for this connection
+    pub fn set_udp_addr(&mut self, addr: SocketAddr) {
+        match addr {
+            SocketAddr::V4(addr) => self.udp_addr_v4 = Some(ConnectionAddr::new(addr)),
```
kpp: https://doc.rust-lang.org/std/option/enum.Option.html#method.replace

kurnevsky: Replace is used to get old value which I don't need. I just need to put a new value there.

kpp: Too bad Option does not have a method like `::set`.

kurnevsky: I think it's because `=` does exact same thing :)

kpp: :D

PR #350, opened 2019-03-23
Add function to kill a crypto connection

kurnevsky:


src/toxcore/net_crypto/mod.rs
```diff
@@ -213,6 +213,37 @@ impl NetCrypto {
...
+
+    /// Kill a connection sending `PACKET_ID_KILL` packet and removing it from
+    /// the connections list.
+    pub fn kill_connection(&self, real_pk: PublicKey) -> impl Future<Item = (), Error = KillConnectionError> {
+        if let Some(connection) = self.connections.write().remove(&real_pk) {
+            let mut connection = connection.write();
+            self.clear_keys_by_addr(&connection);
+            if connection.is_established() || connection.is_not_confirmed() {
+                let packet_number = connection.send_array.buffer_end;
+                Either::A(self.send_data_packet(&mut connection, vec![PACKET_ID_KILL], packet_number)
```
kpp: `vec![PACKET_ID_KILL]` oO I understand that `PacketKill` serializes to `[PACKET_ID_KILL]` but it looks weird and unnatural. Are you sure you want to use this code?

kurnevsky: I'm not sure if we can parse packets here, at this moment they are just arrays. The reason for this is that received packet can be anything: audio, video, message, or even custom packet defined by a client. Maybe later we can use enum here but now we even don't have definitions for all packets we can receive here.

PR #351, opened 2019-03-27
Add "new" method to DhtPkAnnouncePayload

kurnevsky:


src/toxcore/dht/packet/dht_request.rs
```diff
@@ -368,6 +371,17 @@ impl ToBytes for DhtPkAnnouncePayload {
     }
 }
 
+impl DhtPkAnnouncePayload {
+    /// Create new `DhtPkAnnouncePayload` with `no_reply` set to current time.
+    pub fn new(dht_pk: PublicKey, nodes: Vec<TcpUdpPackedNode>) -> Self {
+        DhtPkAnnouncePayload {
```
kpp: use std::time::SystemTime; here

kurnevsky: Then what about `use crate::toxcore::time::*;`?

kpp: Right. That could be included too

PR #356, opened 2019-04-08
Add serde of invite packet for conferences

NamsooCho:


src/toxcore/group/packet/invite.rs
```diff
@@ -0,0 +1,75 @@
+/*! Invite message struct.
```
kpp: > src/toxcore/group/packet/invite.rs

Maybe we should move it into src/toxcore/messenger/packet/group/invite.rs ?

NamsooCho: done

src/toxcore/messenger/packet/group/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Length of group chat unique bytes
+pub const GROUP_UID_BYTES: usize = 32;
+
+/// Unique id used in group chat
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct GroupUID([u8; GROUP_UID_BYTES]);
+
+impl GroupUID {
+    /// Create new object
+    pub fn new() -> Self {
```
kurnevsky: I'd suggest to rename it to `random`, otherwise we will have to add `Default` impl but it doesn't really fit if returns only random values.

kurnevsky: See https://github.com/tox-rs/tox/blob/5d2c2fcd0e32931d294a97389461f8fc710b4744/src/toxcore/tcp/secure.rs#L75

src/toxcore/messenger/packet/group/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+        let mut n = GroupUID([0; GROUP_UID_BYTES]);
+        for (ni, &bsi) in n.0.iter_mut().zip(bs.iter()) {
+            *ni = bsi
+        }
+        Some(n)
+    }
+}
+
+impl FromBytes for GroupUID {
```
kurnevsky: Why there is no `ToBytes`?

kpp: Yeah, why?

src/toxcore/messenger/packet/group/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    );
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn group_uid_from_bytes() {
+        let mut array = [0; GROUP_UID_BYTES];
```
kpp: Also I would like to see standard encode_decode_test tests

src/toxcore/messenger/packet/group/mod.rs
NamsooCho: `GroupUID` is not a packet type. It is not used as an independent type for serde, only used within packets of group chat.
So, I think there is no need to provide `ToBytes` to this type.
This is same to `PublicKey` type, it has only `FromBytes`.

kpp: How can we send it then?

NamsooCho: https://github.com/tox-rs/tox/blob/20ec8d4ae846cd1bd0b9ca125890068a9b00ed36/src/toxcore/messenger/packet/group/invite.rs#L51

src/toxcore/messenger/packet/group/mod.rs
kurnevsky: That's the reason why I asked about `ToBytes` implementation :)

NamsooCho: done

src/toxcore/messenger/mod.rs
```diff
@@ -1,4 +1,6 @@
 /*! The implementation of Messenger
 */
 
-pub mod packet;
+mod packet;
+
+pub use self::packet::*;
```
kpp: I prefer `pub mod packet;` over `mod packet;` + `pub use self::packet::*`;

It's our common style to do so. CHeck out https://github.com/tox-rs/tox/blob/master/src/toxcore/tcp/mod.rs https://github.com/tox-rs/tox/blob/master/src/toxcore/dht/mod.rs#L4 and so on

src/toxcore/messenger/packet/messenger/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+pub use self::typing::*;
+pub use self::user_status::*;
+pub use self::file_data::*;
+pub use self::status_message::*;
+pub use self::file_send_request::*;
+
+/** Messenger packet enum that encapsulates all types of Messenger packets.
+*/
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Packet {
```
kpp: this should be in `messenger/packet/mod.rs`

src/toxcore/messenger/mod.rs
NamsooCho: done

src/toxcore/messenger/packet/messenger/mod.rs
NamsooCho: I changed structure of directory because of this problem https://github.com/tox-rs/tox/pull/358#discussion_r279135935

src/toxcore/messenger/mod.rs
```diff
@@ -2,3 +2,4 @@
 */
 
 pub mod packet;
+pub mod group;
```
kpp: Add newline

NamsooCho: done

kpp:
Another question is how to add group packets to messenger::packet::Packet enum.
Shall we add them grouped by `Group` like this:

```
pub enum Packet {
    ...
    Msi(Msi),
    Group(group::packet::Packet),
}
```

or expanded:

```
pub enum Packet {
    ...
    Msi(Msi),
    GroupInvite(group::packet::Packet::Invide),
    ...
}
```

@kurnevsky  ideas?

NamsooCho:
The bytes stream of Packets are unique. We can distinguish each packets from bytes stream.
One parser can parse all packets of tox-protocol then dispatch packet to its processor.
So, for messenger packet, it can be sent to messenger handler, for group chat packet, it can be sent to group chat handler, for group v2 packet, it can be sent to group v2 handler.
I think there is no need to integrate group chat Packet enum to Messenger Packet enum.

kurnevsky:
I think I'd prefer the latter option since this enum represents the binary form of these packets (in serialized form packets have flat structure).

kpp:
Conclusion:

* `Group` should be renamed to `Conference`.
* We should group `Conference` packets by type to have one `Conference(conference::packet::Packet)` in `enum Messenger::Packet`.

NamsooCho:
>We should group Conference packets by type to have one Conference(conference::packet::Packet) in enum Messenger::Packet.

It will be included in #358

src/toxcore/messenger/conference/mod.rs
```diff
@@ -0,0 +1,4 @@
+/*! Group chat Packets
```
kpp: Group chat?

src/toxcore/messenger/conference/packet/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+mod invite;
+
+pub use self::invite::*;
+
+use nom::be_u8;
+use crate::toxcore::binary_io::*;
+use crate::toxcore::crypto_core::*;
+
+/// Length of conference unique bytes
+pub const GROUP_UID_BYTES: usize = 32;
```
kpp: GROUP?

PR #361, opened 2019-04-13
Update to sodiumoxide alpha-0.3

kpp:


src/toxcore/binary_io.rs
```diff
@@ -101,10 +101,9 @@ macro_rules! encode_decode_test (
     ($test:ident, $value:expr) => (
         #[test]
         fn $test() {
+            #[allow(unused_imports)]
```
kurnevsky: Why not remove this `use`?

kpp: Because we use gen_nonce and other crypto functions inside of this macro

kpp:
@NamsooCho please report whether it worked for you on Mac hardware

PR #362, opened 2019-04-16
WIP: Replace Mutex<u64> with AtomicU64

NamsooCho:


kurnevsky:
AtomicUsize should be replaced with AtomicU64, but not the Mutex. We should check with attr `target_has_atomic` if the target has this atomic and if not - fallback to the mutex.

kpp:
I thought since it is in stable we should not care

kurnevsky:
There might be platforms with no AtomicU64 support. That's why they are behind target_has_atomic flag. I'm not sure what exact platforms have no AtomicU64 (it seems x86 has it?) but either way we should not have both AtomicU64 and AtomicUsize at the same time.

kpp:
Are you sure about that? There is no info about it in the module documentation

kurnevsky:
Check https://github.com/rust-lang/rfcs/blob/master/text/1543-integer_atomics.md#target-support

kpp:
This is a 2016 RFC. I want a little bit fresher info

kurnevsky:
@kpp see module documentation https://doc.rust-lang.org/std/sync/atomic/index.html

> PowerPC and MIPS platforms with 32-bit pointers do not have AtomicU64 or AtomicI64 types.
> ARM platforms like armv5te that aren't for Linux do not have any atomics at all.
> ARM targets with thumbv6m do not have atomic operations at all.

kpp:
Thanks. Then we should fallback to mutexes instead

NamsooCho:
https://github.com/rust-lang/rust/issues/32976#issuecomment-476049327
Issue re-opened.

kurnevsky:
It's fine now but doesn't compile on stable rust. We can either wait when `target_has_atomic` attribute is stabilized or use `target_pointer_width` as before.

kpp:
wait when target_has_atomic attribute is stabilized

kpp:
https://github.com/tox-rs/tox/pull/406

PR #367, opened 2019-04-27
Friends connection

kurnevsky:


src/toxcore/dht/server/mod.rs
```diff
@@ -206,7 +211,7 @@ impl Server {
             sk,
             pk,
             tx,
-            friend_saddr_sink: None,
+            friend_saddr_sink: Arc::new(RwLock::new(None)),
```
kpp: Default::default()?

src/toxcore/dht/server/mod.rs
```diff
@@ -1461,6 +1492,12 @@ mod tests {
     const ONION_RETURN_2_PAYLOAD_SIZE: usize = ONION_RETURN_2_SIZE - secretbox::NONCEBYTES;
     const ONION_RETURN_3_PAYLOAD_SIZE: usize = ONION_RETURN_3_SIZE - secretbox::NONCEBYTES;
 
+    impl Server {
+        pub fn has_friend(&self, pk: &PublicKey) -> bool {
+            self.friends.read().contains_key(pk)
```
kpp: I doubt whether dht::Server should not answer whether it has friend or not.

src/toxcore/friend_connection/mod.rs
```diff
@@ -1,4 +1,1070 @@
...
+                        info!("Found a friend's IP address");
+
+                        friend.saddr = Some(node.saddr);
+
+                        net_crypto.add_connection(friend.real_pk, node.pk);
+                        net_crypto.set_friend_udp_addr(friend.real_pk, node.saddr);
+                    }
+                }
+
+                future::ok(())
```
kpp: Ok(())?

src/toxcore/friend_connection/mod.rs
```diff
@@ -1,4 +1,1070 @@
...
+    }
+
+    /// Handle the stream of found IP addresses.
+    fn handle_friend_saddr(&self, friend_saddr_rx: mpsc::UnboundedReceiver<PackedNode>) -> impl Future<Item = (), Error = RunError> + Send {
+        let net_crypto = self.net_crypto.clone();
+        let friends = self.friends.clone();
+        friend_saddr_rx
+            .map_err(|()| unreachable!("rx can't fail"))
+            .for_each(move |node| {
+                if let Some(friend) = friends.write().values_mut().find(|friend| friend.dht_pk == Some(node.pk)) {
```
kpp: If `friend: Copy` you will make a shallow copy in `if let Some(friend)`

src/toxcore/dht/server/mod.rs
kurnevsky: What do you mean?

src/toxcore/friend_connection/mod.rs
kurnevsky: It doesn't implement Copy, and even if it were it'd still be a reference.

src/toxcore/dht/server/mod.rs
kpp: I think it's a responsibility of Messenger, not dht::Server.

kurnevsky: But it's for test purposes only - it's located inside `tests` module. I use it to check if friends module really adds/removes friends into/from internal modules.

kpp: Oh sorry

PR #371, opened 2019-05-10
Re-grouping file transfer packets from messenger

NamsooCho:


src/toxcore/messenger/file_transfer/packet/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        randombytes_into(&mut array);
+        FileUID(array)
+    }
+
+    fn from_slice(bs: &[u8]) -> Option<FileUID> {
+        if bs.len() != FILE_UID_BYTES {
+            return None
+        }
+        let mut n = FileUID([0; FILE_UID_BYTES]);
+        for (ni, &bsi) in n.0.iter_mut().zip(bs.iter()) {
```
kpp: https://doc.rust-lang.org/stable/std/primitive.slice.html#method.clone_from_slice

NamsooCho: done

PR #372, opened 2019-05-14
migrate to nom5 on PingRequest

NamsooCho:
Currently, nom 3.2 and nom 5 are used at the same time.

### DIFF
--- a/src/toxcore/dht/packet/errors.rs
+++ b/src/toxcore/dht/packet/errors.rs
@@ -33,6 +33,10 @@ impl GetPayloadError {
     pub(crate) fn deserialize(error: NomErrorKind, payload: Vec<u8>) -> GetPayloadError {
         GetPayloadError::from(GetPayloadErrorKind::Deserialize { error, payload })
     }
+
+    pub(crate) fn deserialize5(error: nom5::error::ErrorKind, payload: Vec<u8>) -> GetPayloadError {
+        GetPayloadError::from(GetPayloadErrorKind::Deserialize5 { error, payload })
+    }
 }
 
 impl Fail for GetPayloadError {
@@ -70,6 +74,14 @@ pub enum GetPayloadErrorKind {
     Deserialize {
         /// Parsing error
         error: NomErrorKind,
+        /// Received payload of packet
+        payload: Vec<u8>,
+    },
+    /// Error indicates that decrypted payload of packet can't be parsed, it is for nom 5.
+    #[fail(display = "Deserialize payload error: {:?}, data: {:?}", error, payload)]
+    Deserialize5 {
+        /// Parsing error
+        error: nom5::error::ErrorKind,
         /// Received payload of packet
         payload: Vec<u8>,
     }
--- a/src/toxcore/dht/packet/ping_request.rs
+++ b/src/toxcore/dht/packet/ping_request.rs
@@ -3,6 +3,14 @@
 
 use nom::{be_u64, rest};
 
+use nom5::{IResult, Needed, Err,
+    bytes::complete::tag,
+    number::complete::be_u64 as be_u64_5,
+    sequence::{delimited, tuple},
+    combinator::{map, rest as rest_5},
+    error::ParseError
+};
+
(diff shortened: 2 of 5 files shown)


src/toxcore/dht/packet/ping_request.rs
```diff
@@ -72,6 +86,28 @@ impl PingRequest {
             payload,
         }
     }
+
+    // Root parser for PingRequest object
+    fn parse<'a, E: ParseError<&'a [u8]>>(i: &'a [u8]) ->IResult<&'a [u8], PingRequest, E> {
```
kpp: This function is very verbose. We used nom3's do_parse to make code compact. I would like you to implement the same in nom5.

src/toxcore/dht/packet/errors.rs
```diff
@@ -72,6 +76,14 @@ pub enum GetPayloadErrorKind {
         error: NomErrorKind,
         /// Received payload of packet
         payload: Vec<u8>,
+    },
+    /// Error indicates that decrypted payload of packet can't be parsed, it is for nom 5.
+    #[fail(display = "Deserialize payload error: {:?}, data: {:?}", error, payload)]
+    Deserialize5 {
+    /// Parsing error
```
kurnevsky: Wrong padding.

src/toxcore/dht/packet/ping_request.rs
```diff
@@ -150,12 +198,31 @@ impl ToBytes for PingRequestPayload {
     }
 }
 
+impl PingRequestPayload {
+    // Root parser for PingRequest object
+    fn parse<'a, E: ParseError<&'a [u8]>>(i: &'a [u8]) ->IResult<&'a [u8], PingRequestPayload, E> {
```
kurnevsky: Why is it a separate function? Why not just have this parsing inside `from_bytes5`?

src/toxcore/dht/packet/ping_request.rs
kurnevsky: It seems it can be simplified to:

```rust
map(tuple((
    tag("\x00"),
    PublicKey::from_bytes5,
    Nonce::from_bytes5,
    map(rest_5, |bytes: &[u8]| bytes.to_vec()),
)), |(_, pk, nonce, payload)| PingRequest { pk, nonce, payload })
```

What do you think about this option? Honestly I'd prefer plain functions rather than using weird looking dsl (which also can't be formatted with rustfmt).

kurnevsky: `do_parse` is still there https://docs.rs/nom/5.0.0-beta1/nom/macro.do_parse.html
So nevermind, I'm fine using it - it should be easier to migrate.

src/toxcore/dht/packet/ping_request.rs
NamsooCho: done

src/toxcore/dht/packet/ping_request.rs
```diff
@@ -48,14 +66,15 @@ impl ToBytes for PingRequest {
...
-    named!(from_bytes<PingRequest>, do_parse!(
-        tag!("\x00") >>
-        pk: call!(PublicKey::from_bytes) >>
-        nonce: call!(Nonce::from_bytes) >>
-        payload: map!(rest, |bytes| bytes.to_vec() ) >>
-        (PingRequest { pk, nonce, payload })
-    ));
+impl FromBytes5 for PingRequest {
+    fn from_bytes5<'a, E: ParseError<&'a [u8]>>(i: &'a [u8]) ->IResult<&'a [u8], PingRequest, E> {
+        map(tuple((
```
kpp: Namsoo, in the best case we should migrate to nom5 only by changing the version of `nom` in out Cargo.toml with almost no changes.

So the code here should be as close to `FromBytes` as possible. With macro named and do_parse.

kpp: If these macros are missing in nom5 you have to find a way to implement them

NamsooCho:
In favor of #374, close this.

PR #376, opened 2019-05-18
Update to sodiumoxide 0.2.2

kpp:
@NamsooCho please run tests on your Mac. Remember you added explicit crypto_init() in tests because they fail on your Mac? 2576fcb05a8b15a06

### DIFF
--- a/src/toxcore/crypto_core.rs
+++ b/src/toxcore/crypto_core.rs
@@ -11,14 +11,24 @@ use crate::toxcore::binary_io::*;
 
 // TODO: check if `#[inline]` is actually useful
 
+/** Run before using crypto.
 
-/// Initialize sodiumoxide
-#[deprecated(
-    since = "0.0.10",
-    note = "libsodium is automatically initialized by sodiumoxide now."
-)]
+Runs [`sodiumoxide::init()`](../../../sodiumoxide/fn.init.html).
+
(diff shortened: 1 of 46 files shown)


src/toxcore/crypto_core.rs
```diff
@@ -226,6 +236,32 @@ impl FromBytes for sha512::Digest {
...
     use super::*;
 
+    use std::thread;
+
+    // crypto_init()
+
+    #[test]
+    // test whether crypto init behaves well when one tries to run it in parallel.
+    // Its code should **not** be ran in parallel.
+    fn crypto_init_test() {
```
kurnevsky: Do we need this test? It looks like it should be tested/guaranteed by sodiumoxide.

kpp: No it's not. That feature was removed right before release.

See
 https://github.com/sodiumoxide/sodiumoxide/pull/336
 https://github.com/sodiumoxide/sodiumoxide/blob/master/src/lib.rs#L93

kurnevsky: I know, and that's not what I meant. This test is meant to check thread safety for `crypto_init` but it's not our responsibility to make it thread safe. It should be guaranteed by sodiumoxide,

kpp: I see. You are right

PR #380, opened 2019-05-28
Handle friend requests inside onion client

kurnevsky:


src/toxcore/io_tokio.rs
```diff
@@ -32,3 +32,51 @@ pub fn send_all_to<T: Send + 'static, S, Tx, E: Debug>(tx: &Tx, s: S) -> impl Fu
         .map(|_tx| ()) // ignore tx because it was cloned
 }
 
+/// `Sink` type that can either be empty or store inner `Sink`. When it's empty
+/// it drops all sent elements.
+#[derive(Debug, Clone)]
```
kpp: derive Default,

pub fn new() will be: { Default::default() }

kurnevsky: Nope, it won't work. `derive` requires trait implementation for all generic type parameters. In our case it will require `Default` for senders which they don't implement.

kpp: Did you try?
https://doc.rust-lang.org/src/core/option.rs.html#1042-1046

kpp: Looks like a bug. I will file it to rustlang.

src/toxcore/io_tokio.rs
```diff
@@ -32,3 +32,51 @@ pub fn send_all_to<T: Send + 'static, S, Tx, E: Debug>(tx: &Tx, s: S) -> impl Fu
...
+pub struct OptionalSink<S: Sink>(Option<S>);
+
+impl<S: Sink> OptionalSink<S> {
+    /// Create new `OptionalSink`.
+    pub fn new() -> Self {
+        OptionalSink(None)
+    }
+
+    /// Put a sink to `OptionalSink`.
+    pub fn put(&mut self, sink: S) {
```
kpp: set?

src/toxcore/io_tokio.rs
kurnevsky: https://stackoverflow.com/questions/46079916/possible-bug-in-generic-implementation-of-default

src/toxcore/io_tokio.rs
```diff
@@ -32,3 +32,51 @@ pub fn send_all_to<T: Send + 'static, S, Tx, E: Debug>(tx: &Tx, s: S) -> impl Fu
         .map(|_tx| ()) // ignore tx because it was cloned
 }
 
+/// `Sink` type that can either be empty or store inner `Sink`. When it's empty
+/// it drops all sent elements.
+#[derive(Debug, Clone)]
+pub struct OptionalSink<S: Sink>(Option<S>);
+
+impl<S: Sink> OptionalSink<S> {
```
kpp: This PR looks good except I don't like OptionalSink in io_tokio.rs file. Move it somewhere

kurnevsky: OptionalSink and send_to/send_all_to have common topic - they both are auxiliary utils for futures. Also we are getting rid of `IoFuture`/`IoStream` in favor of `impl Future` and async/await in the future. So probably we can remain it together and just rename `io_rokio` mod somehow?

kpp: Fine, let think about it later

PR #383, opened 2019-05-30
Search requests should contain zero pind_id

kurnevsky:


src/toxcore/onion/client/mod.rs
```diff
@@ -658,7 +658,14 @@ impl OnionClient {
...
 
-                let inner_announce_request = announce_packet_data.request(&node.pk, node.ping_id, request_id);
+                // the request is a search request if pind_id is 0 and an
+                // announce request otherwise
+                let ping_id = if friend_pk.is_some() {
+                    None
+                } else {
+                    node.ping_id
+                };
+                let inner_announce_request = announce_packet_data.request(&node.pk, ping_id, request_id);
```
kpp: how about adding different methods to call like: `announce_packet_data.search_request` or `announce_packet_data.announce_request`?

PR #389, opened 2019-06-12
Add TCP support for onion client

kurnevsky:


src/toxcore/dht/server/mod.rs
```diff
@@ -1492,6 +1499,10 @@ mod tests {
         pub fn has_friend(&self, pk: &PublicKey) -> bool {
             self.friends.read().contains_key(pk)
         }
+
+        pub fn add_node(&self, node: PackedNode) {
+            assert!(self.close_nodes.write().try_add(node));
```
kpp: assert! is not a good idea since a debug version can fail in runtime and crash the app.

kurnevsky: `add_node` is for tests only.

kpp: ><' Sorry. github does not always show the context correctly.

src/toxcore/onion/client/mod.rs
```diff
@@ -2301,4 +2348,64 @@ mod tests {
...
+            nonce: gen_nonce(),
+            pk: gen_keypair().0,
+            payload: vec![42; 123],
+        });
+
+        let path = OnionPath::new([
+            PackedNode::new("127.0.0.1:12346".parse().unwrap(), &relay_pk),
+            PackedNode::new("127.0.0.1:12347".parse().unwrap(), &gen_keypair().0),
+            PackedNode::new("127.0.0.1:12348".parse().unwrap(), &gen_keypair().0),
+        ], /* TCP */ true);
```
kpp: Please add TODO to replace bool with enum

src/toxcore/onion/client/nodes_pool.rs
```diff
@@ -52,6 +56,47 @@ impl NodesPool {
...
+        loop {
+            node_2 = self.rand().unwrap();
+            if node_2 != node_1 {
+                break;
+            }
+        }
+        let mut node_3;
+        loop {
+            node_3 = self.rand().unwrap();
+            if node_3 != node_1 && node_3 != node_2 {
```
kpp: What if there is less than 3 nodes in the pool? Will you enter an infinite loop?

src/toxcore/onion/client/nodes_pool.rs
```diff
@@ -52,6 +56,47 @@ impl NodesPool {
     pub fn len(&self) -> usize {
         self.nodes.len()
     }
+
+    /// Build new random onion path with first UDP node.
+    pub fn udp_path(&self) -> Option<OnionPath> {
```
kpp: Is there a way to shuffle a Vec and take first three items?

kurnevsky: Not with sodiumoxide being used for rng. Also shuffling might be less performant nodes pool is filled.

src/toxcore/onion/client/mod.rs
kpp: Or a new issue

kurnevsky: Simply added enum :)

PR #390, opened 2019-06-14
refactor(friends) remove hardcoded keys in example/echo.rs

### LINKED ISSUE #379: examples/echo.rs: remove hardcoded keys
Generate an id each time and take friend's TOX_ID from cli args

inosms:
resolves #379 
I completely removed the hardcoded keys and the current implementation just accepts all friend requests (ignoring non spam, etc.)

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -23,6 +23,7 @@ use tox::toxcore::net_crypto::{NetCrypto, NetCryptoNewArgs};
 use tox::toxcore::onion::client::OnionClient;
 use tox::toxcore::tcp::client::Connections;
 use tox::toxcore::stats::Stats;
+use tox::toxcore::toxid::ToxId;
 
 const BOOTSTRAP_NODES: [(&str, &str); 9] = [
     // Impyy
@@ -45,10 +46,6 @@ const BOOTSTRAP_NODES: [(&str, &str); 9] = [
     ("2B2137E094F743AC8BD44652C55F41DFACC502F125E99E4FE24D40537489E32F", "5.189.176.217:5190"),
 ];
 
-const SELF_SK: &str = "1A5EC1D6C3F1FA720A313C01F432B6AE0D4649A5121964C9992DDF32871E8DFD";
-
-const FRIEND_PK: &str = "3E6A06DA48D1AB98549AD76890770B704AE9116D8654FBCD35C9BF2DB9233E21";
-
 /// Bind a UDP listener to the socket address.
 fn bind_socket(addr: SocketAddr) -> UdpSocket {
     let socket = UdpSocket::bind(&addr).expect("Failed to bind UDP socket");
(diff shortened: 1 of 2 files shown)


examples/echo.rs
```diff
@@ -163,6 +158,15 @@ fn main() {
             }
         });
 
+    // handle incoming friend connections by just accepting all of of them
+    let friend_connection_c_2 = friend_connections.clone();
```
kurnevsky: In rust it's ok to have several `let`s with the same name, so we can call it just `friend_connections_c`. Also let it be plural to be consistent with other names.

examples/echo.rs
```diff
@@ -163,6 +158,15 @@ fn main() {
...
+    let friend_future = friend_request_sink_rx
+        .map_err(|()| unreachable!("rx can't fail"))
+        .for_each(move |(pk, _)| {
+            friend_connection_c_2.add_friend(pk);
+            Box::new(future::ok(()))
```
kurnevsky: No need to put it into `Box`.

examples/echo.rs
```diff
@@ -176,6 +180,7 @@ fn main() {
         Box::new(friend_connections.run().map_err(Error::from)),
         Box::new(lossless_future),
         Box::new(lossy_future),
+        Box::new(friend_future)
```
kurnevsky: It makes sense to put trailing comma here so that if we need to add one more future we will change only one line - that allows to have more precise git blames.

examples/echo.rs
```diff
@@ -163,6 +158,15 @@ fn main() {
             }
         });
 
+    // handle incoming friend connections by just accepting all of of them
```
kurnevsky: friend connections -> friend requests
of of -> of

examples/echo.rs
```diff
@@ -91,6 +88,9 @@ fn main() {
     let (lossless_tx, lossless_rx) = mpsc::unbounded();
     let (lossy_tx, lossy_rx) = mpsc::unbounded();
 
+    let (friend_request_sink_tx, friend_request_sink_rx) = mpsc::unbounded();
```
kurnevsky: The word `sink` is redundant here. It's a (sink, stream) tuple or in other words (tx, rx).

examples/echo.rs
```diff
@@ -64,9 +60,10 @@ fn main() {
 
     let (dht_pk, dht_sk) = gen_keypair();
 
-    let real_sk_bytes: [u8; 32] = FromHex::from_hex(SELF_SK).unwrap();
-    let real_sk = SecretKey::from_slice(&real_sk_bytes).unwrap();
-    let real_pk = real_sk.public_key();
+    // create random tox id and print it
+    let (real_pk, real_sk) = gen_keypair();
+    let id = tox::toxcore::toxid::ToxId::new(real_pk.clone());
```
kurnevsky: Maybe import it by adding `use tox::toxcore::toxid::ToxId` to the header?

examples/echo.rs
```diff
@@ -64,9 +60,10 @@ fn main() {
...
+    println!("your tox id is: {:X}\n",id);
```
kurnevsky: `println` inserts new line at the end itself, so it will print one extra new line because of `\n`.

inosms:
I felt so free to squash both commits as I thought the second one doesn't need to reappear in the master commit history

PR #392, opened 2019-06-20
Handle TCP net_crypto packets

kurnevsky:


src/toxcore/onion/client/mod.rs
```diff
@@ -1650,7 +1650,7 @@ mod tests {
             payload: vec![42; 123],
         };
 
-        let error = onion_client.handle_announce_response(&packet, saddr).wait().err().unwrap();
+        let error = onion_client.handle_announce_response(&packet, true).wait().err().unwrap();
```
kpp: `handle_announce_response(&packet, true)` does not look great. Please add a enum for `Global/Local` IPs

kurnevsky: It's a test and the only place where we need to pass true is handling tcp packets. When we handle udp packet it's calculated from address like `IsGlobal::is_global(addr)`. So if I change it to enum I'll have to add `if IsGlobal::is_global {Enum::Global} else {Enum::Local}` there. Do you think it worth it?

suhr: Maybe there should be `Locality` trait instead with a method that returns `Global` or `Local`.

kpp: Does not matter. https://github.com/rust-lang/rust/issues/27709#issuecomment-556459434

PR #395, opened 2019-08-04
Add explicit dyn keyword

kpp:


src/toxcore/dht/packet/errors.rs
```diff
@@ -36,7 +36,7 @@ impl GetPayloadError {
 }
 
 impl Fail for GetPayloadError {
-    fn cause(&self) -> Option<&Fail> {
+    fn cause(&self) -> Option<&dyn Fail> {
```
kpp: @kurnevsky did you miss that one?

kurnevsky: Yes it seems...

PR #406, opened 2019-10-03
refactor(stats): Use AtomicU64 instead of locks

### LINKED ISSUE #360: Replace stats::mutex with atomic u64
https://github.com/tox-rs/tox/blob/5d2c2fcd0e32931d294a97389461f8fc710b4744/src/toxcore/stats.rs#L41

https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html is in stable

cronokirby:
Fixes #360

Previously, we wanted to use `AtomicUsize` as an atomic type large enough for a 64 bit counter. This meant having two versions of the code, one with atomics, another with locks. With the introduction of `AtomicU64` into stable, we can simply use this type no matter what.

This removes code duplication, and allows us to avoid using locks on platforms with 32 bit usize.

### DIFF
--- a/src/toxcore/stats.rs
+++ b/src/toxcore/stats.rs
@@ -23,73 +23,34 @@ impl Stats {
     }
 }
 
-#[cfg(target_pointer_width = "64")]
-#[derive(Default)]
-/// Struct for counting packets on x64 CPU.
-pub struct Counters {
-    /// Incoming packets count for Udp/Tcp
-    incoming: AtomicUsize,
-    /// Outgoing packets count for Udp/Tcp
-    outgoing: AtomicUsize,
-}
-
-#[cfg(not(target_pointer_width = "64"))]
 #[derive(Default)]
 /// Struct for counting packets on non-x64 CPU.
 pub struct Counters {
     /// Incoming packets count for Udp/Tcp
-    incoming: Mutex<u64>,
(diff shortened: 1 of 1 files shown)


src/toxcore/stats.rs
```diff
@@ -23,73 +23,34 @@ impl Stats {
...
-pub struct Counters {
-    /// Incoming packets count for Udp/Tcp
-    incoming: AtomicUsize,
-    /// Outgoing packets count for Udp/Tcp
-    outgoing: AtomicUsize,
-}
-
-#[cfg(not(target_pointer_width = "64"))]
 #[derive(Default)]
 /// Struct for counting packets on non-x64 CPU.
```
kpp: > non-x64 CPU.

Should be removed

src/toxcore/stats.rs
```diff
@@ -23,73 +20,33 @@ impl Stats {
     }
 }
 
-#[cfg(target_pointer_width = "64")]
 #[derive(Default)]
-/// Struct for counting packets on x64 CPU.
```
kpp: Take a look at warnings: https://travis-ci.com/tox-rs/tox/jobs/241932457#L382

At least this should be smt like: `/// Struct for counting packets` to please the compiler

PR #420, opened 2020-02-22
Split toxcore into separate crates

suhr:


binary_io/src/lib.rs
```diff
@@ -1,24 +1,42 @@
...
-*/
-
 pub use nom::IResult;
 pub use cookie_factory::GenError;
 
 use nom::number::streaming::{le_u8, le_u16};
+use nom::{named, map, count};
+use cookie_factory::{do_gen, gen_be_u8, gen_le_u16};
+
+#[cfg(feature = "sodiumoxide")]
```
kurnevsky, line 8: What's the purpose of this feature? You use it in almost every line, so it might be better to move all this to a separate module and define it with this feature.

tox_core/src/state_format/old.rs
```diff
@@ -615,14 +615,16 @@ impl ToBytes for State {
 mod tests {
     use super::*;
 
-    use crate::toxcore::ip_port::*;
+    use tox_packet::ip_port::*;
 
     encode_decode_test!(
+        tox_crypto::crypto_init().unwrap(),
```
kurnevsky, line 621: Is it necessary?

binary_io/src/lib.rs
suhr: Orphan rules. You can't implement a foreign trait for a foreign type.

tox_packet/src/onion/onion_data_response.rs
```diff
@@ -375,19 +371,19 @@ mod tests {
...
+    //     let nonce = gen_nonce();
+    //     let shared_secret = encrypt_precompute(&bob_pk, &alice_sk);
+    //     let friend_request = FriendRequest::new(NoSpam::random(), "1234".to_owned());
+    //     let payload = OnionDataResponseInnerPayload::FriendRequest(friend_request);
+    //     // encode payload
+    //     let packet = OnionDataResponsePayload::new(&shared_secret, alice_pk, &nonce, &payload);
+    //     // decode payload
+    //     let decoded_payload = packet.get_payload(&nonce, &shared_secret).unwrap();
+    //     assert_eq!(payload, decoded_payload);
+    // }
```
kurnevsky, line 388: Why commented?

binary_io/src/lib.rs
kurnevsky: Aren't they per crate?

Cargo.toml
```diff
@@ -1,60 +1,11 @@
...
-version = "0.0.10"
-authors = [
-  "Zetok Zalbavar <zetok@openmailbox.org>",
-  "Roman Proskuryakov <humbug@deeptown.org>",
-  "Namsoo CHO <nscho66@gmail.com>",
-  "Evgeny Kurnevsky <kurnevsky@gmail.com>",
+[workspace]
+members = [
+    "tox",
+    "binary_io",
```
kurnevsky, line 4: Should it be prefixed with `tox_`?

suhr: It's not really tox specific, though I can prefix it with `tox_` if necessary.

tox_packet/src/onion/onion_data_response.rs
suhr: Indeed, it should not.

Cargo.toml
kurnevsky: It'll have this name on cartes.io which might be confusing, especially if it contains tox-specific encoding for `IpAddr`.

tox_core/src/state_format/old.rs
suhr: Sodiumoxide still needs an initialization AFAIK.

binary_io/src/lib.rs
suhr: Indeed.

tox_core/src/state_format/old.rs
kurnevsky: But we shouldn't use sodiumoxide in such tests - it's just nom+cookie_factory.

kurnevsky: Ah, no, we generate pk/nonce.

PR #425, opened 2020-05-10
Use idiomatic sodiumoxide API for password KDF

unseddd:
Uses the sodiumoxide API from the [documentation example](https://docs.rs/sodiumoxide/0.2.5/sodiumoxide/crypto/pwhash/index.html#example-key-derivation) for deriving
a key from a passphrase, and securely zeroing the intermediate key.

### DIFF
--- a/tox_encryptsave/src/lib.rs
+++ b/tox_encryptsave/src/lib.rs
@@ -33,15 +33,13 @@ use tox_crypto::{
     NONCEBYTES, MACBYTES,
     Nonce, PrecomputedKey,
     gen_nonce,
-    sha256, memzero
+    secretbox, sha256
 };
 
 /// Length in bytes of the salt used to encrypt/decrypt data.
 pub use tox_crypto::pwhash::SALTBYTES as SALT_LENGTH;
 /// Length in bytes of the key used to encrypt/decrypt data.
 pub use tox_crypto::PRECOMPUTEDKEYBYTES as KEY_LENGTH;
-
-use tox_crypto;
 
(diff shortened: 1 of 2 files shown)


tox_encryptsave/src/lib.rs
```diff
@@ -33,7 +33,7 @@ use tox_crypto::{
     NONCEBYTES, MACBYTES,
     Nonce, PrecomputedKey,
     gen_nonce,
-    sha256, memzero
```
kpp, line 36: Do we use memzero somewhere else?

suhr: No.

kpp: @ unseddd how about removing memzero from tox_crypto if we dont use it?

unseddd: > @unseddd how about removing memzero from tox_crypto if we dont use it?

No problem, wanted to make minimal changes. Can remove `memzero` from the rest of the crate.

Will fix my commit message with the update :)

suhr:
Hm, clippy is angry: https://github.com/tox-rs/tox/pull/425/checks?check_run_id=660913040#step:5:161

unseddd:
Could only find a warning about `use tox_crypto;` being a redundant import.

Removed it, and clippy + tests all pass. If the clippy test fails again, could you point me to the checks it is performing?

unseddd:
Could only find a warning about use tox_crypto; being a redundant import.

Looks like that made clippy happy :smile:

Issue #414, opened 2019-11-01
Move to async/await


kpp:
- [x] migrate to futures 0.3 and tokio 0.2
- [x] replace all `fn -> impl Future` with `async fn`

kurnevsky:
Was done in https://github.com/tox-rs/tox/pull/434

PR #426, opened 2020-05-12
[WIP] Use async/await semantics in DHT server

unseddd:
Change `impl Future` function return types to async/await semantics.
Use modern syntax for asynchronous coding.

Related to #414

### DIFF
--- a/tox_core/src/onion/client/mod.rs
+++ b/tox_core/src/onion/client/mod.rs
@@ -533,16 +533,16 @@ impl OnionClient {
     }
 
     /// Handle DHT `PublicKey` announce from both onion and DHT.
-    pub fn handle_dht_pk_announce(&self, friend_pk: PublicKey, dht_pk_announce: DhtPkAnnouncePayload) -> impl Future<Output = Result<(), HandleDhtPkAnnounceError>> + Send {
-        let mut state = self.state.lock();
+    pub async fn handle_dht_pk_announce(&self, friend_pk: PublicKey, dht_pk_announce: DhtPkAnnouncePayload) -> Result<(), HandleDhtPkAnnounceError> {
+        let mut state = self.state.lock().clone();
 
         let friend = match state.friends.get_mut(&friend_pk) {
             Some(friend) => friend,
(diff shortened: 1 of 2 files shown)


unseddd:
Changed a test in `tox_core/src/dht/server/mod.rs:2160` because it was `unwrap`ing an error.

Can debug and change if the result of the call on that line should be `Ok(())`. Was not obvious to me that it should have a valid result, since the next lines test for an expected error.

tox_core/src/onion/client/mod.rs
```diff
@@ -533,16 +533,16 @@ impl OnionClient {
     }
 
     /// Handle DHT `PublicKey` announce from both onion and DHT.
-    pub fn handle_dht_pk_announce(&self, friend_pk: PublicKey, dht_pk_announce: DhtPkAnnouncePayload) -> impl Future<Output = Result<(), HandleDhtPkAnnounceError>> + Send {
-        let mut state = self.state.lock();
+    pub async fn handle_dht_pk_announce(&self, friend_pk: PublicKey, dht_pk_announce: DhtPkAnnouncePayload) -> Result<(), HandleDhtPkAnnounceError> {
+        let mut state = self.state.lock().clone();
```
kurnevsky, line 537: Why does it become clone? This way it's not atomic update anymore.

unseddd: Tried using `Copy`, but compiler complains that the lifetime extends past the `await` call below.

I agree, this is hacky. Can you see an obviously better way to do this? Will keep hacking on it for a better solution.

Marking this PR as [WIP] until then.

unseddd: @kurnevsky after wrestling with this for a few more hours, the best thing I could half-way get working ended up looking very similar to what is there now (with `impl Future<Output...>` return).

Am sure there is a way to accomplish the refactor, but maybe it takes a total refactor of all the async components.

Will re-attempt the refactor when more familiar with the library, and with async/await in general. Apologies if I wasted any of your time.

PR #428, opened 2020-06-22
Drop MIT license

kurnevsky:
Making PR to start discussion on the license we use.

One of the reasons why we need to do it is that `tox-rs` is a derived work of `c-toxcore` as it borrows many algorighms/approaches from `c-toxcore` and therefore has to inherit its license.


kpp:
How about GPL with permissions to post the code onto App Store?

kurnevsky:
Should we get an approval from toktok team for that? Ping @zugz

iphydf:
No need for approval. This action is legal.

kurnevsky:
Pushed a note about Apple App Store. The text is taken from https://github.com/nextcloud/ios/blob/master/COPYING.iOS

kpp:
Thanks. We need to fix clippy warnings (in another PR probably).

PR #436, opened 2020-09-20
feat(ktree): implement k-tree with force-k modification

kurnevsky:
Unlike k-tree it holds additional 8 nodes that are always closest to
own PK. It forces a peer to always accept nodes that are closer than
known ones which improves search time.

### DIFF
--- a/tox_core/src/dht/forced_ktree.rs
+++ b/tox_core/src/dht/forced_ktree.rs
@@ -1,117 +1,95 @@
-//! K-buckets structure
+use std::cmp::Ordering;
+
+use itertools::Itertools;
 
 use tox_crypto::*;
-use crate::dht::dht_node::*;
 use tox_packet::dht::packed_node::*;
-use crate::dht::ip_port::IsGlobal;
+use crate::dht::dht_node::*;
 use crate::dht::kbucket::*;
+use crate::dht::ktree::*;
+use crate::dht::ip_port::IsGlobal;
 
 /** K-buckets structure to hold up to
-[`KBUCKET_MAX_ENTRIES`](./constant.KBUCKET_MAX_ENTRIES.html) *
+([`KBUCKET_MAX_ENTRIES`](./constant.KBUCKET_MAX_ENTRIES.html) + 1) *
 [`KBUCKET_DEFAULT_SIZE`](./constant.KBUCKET_DEFAULT_SIZE.html) nodes close to
 own PK.
 
 Buckets in ktree are sorted by closeness to the PK; closest bucket is the last
 one, while furthest is the first one.
 
-Further reading: [Tox spec](https://zetok.github.io/tox-spec#k-buckets).
+This structure implements a force-k modification described in the whitepaper
+"Improving the Performance and Robustness of Kademlia-based Overlay Networks".
 
-The name references to the kademlia binary tree from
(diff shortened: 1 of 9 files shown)


tox_core/src/dht/forced_ktree.rs
```diff
@@ -0,0 +1,438 @@
...
+    Returns less than `count` nodes only if `Ktree` contains less than `count`
+    nodes.
+
+    It should not contain LAN ip node if the request is from global ip.
+    */
+    pub fn get_closest(&self, pk: &PublicKey, count: u8, only_global: bool) -> Kbucket<PackedNode> {
+        let mut kbucket = self.ktree.get_closest(pk, count, only_global);
+        for node in self.kbucket.iter().filter(|node| !node.is_bad()) {
+            if let Some(pn) = node.to_packed_node() {
+                if !only_global || IsGlobal::is_global(&pn.saddr.ip()) {
```
kpp, line 108: Is it a correct check?

tox_core/src/dht/forced_ktree.rs
```diff
@@ -0,0 +1,438 @@
...
+    and node is not already in the [`Bucket`].
+
+    Otherwise `false` is returned.
+
+    [`Bucket`]: ./struct.Bucket.html
+    [`PackedNode`]: ./struct.PackedNode.html
+    */
+    pub fn can_add(&self, new_node: &PackedNode) -> bool {
+        self.ktree.can_add(new_node) ||
+            !self.ktree.contains(&new_node.pk) && self.kbucket.can_add(&self.ktree.pk, new_node, /* evict */ true)
```
kpp, line 139: Is this is a correct check?

tox_core/src/dht/forced_ktree.rs
kurnevsky: Yes, it's either global and we add it, or local and we include not only global so we still add it.

tox_core/src/dht/forced_ktree.rs
kurnevsky: Yes, we don't want to add node twice, so either it can be added to internal ktree (and removed from self if it's there) or to self.

kpp: I mean can't we simplify to just "can_add || can_add"?

kurnevsky: We can't because if ktree can't add a node it can either be full or contain the node. If it's just full we might add the node to self kbucket, but if it already contains the node we don't won't to add it second time.

kpp: That's right but kbucket has `, /* evict */ true` param whereas ktree hasn't. I wonder whether it worth checking that kbucket does not contain the node.

kurnevsky: We check that ktree doesn't contain it. It worth, otherwise it can be added twice.

kpp: However with `self.kbucket.can_add(&self.ktree.pk, new_node, /* evict */ true)` it can be added twice

kurnevsky: No, it can't. It will overwrite previous added node. But if we don't check it it will be literally added twice, so we will send same packets to the same node twice.

kpp (approved):

PR #449, opened 2021-04-26
refactor(sha): use sha2 crate for hashes instead of sodiumoxide

kurnevsky:


tox_core/src/onion/client/mod.rs
```diff
@@ -169,7 +171,7 @@ struct OnionNode {
     /// Path used to send packets to this node.
     path_id: OnionPathId,
     /// Ping id that should be used to announce to this node.
-    ping_id: Option<sha256::Digest>,
+    ping_id: Option<[u8; <Sha256 as Digest>::OutputSize::USIZE]>,
```
kpp, line 174: Would you please add an alias for ping_id type?

tox_packet/src/dht/cookie.rs
```diff
@@ -160,10 +162,11 @@ impl EncryptedCookie {
         }
     }
     /// Calculate SHA512 hash of encrypted cookie together with nonce
-    pub fn hash(&self) -> sha512::Digest {
+    pub fn hash(&self) -> [u8; <Sha512 as Digest>::OutputSize::USIZE] {
         let mut buf = [0; 112];
         let (_, size) = self.to_bytes((&mut buf, 0)).unwrap();
-        sha512::hash(&buf[..size])
+        // TODO: use `Into` directly when GenericArray supports it
```
kpp, line 168: no need to, since generic array will be removed because of const generics

kurnevsky: It's unlikely - const generics are less powerful. It could be [replaced by another library](https://github.com/fizyk20/generic-array/issues/115#issuecomment-808272099) but until then worth to have this todo, otherwise after migration wi might miss this useless transformation.

tox_core/src/onion/client/mod.rs
kurnevsky: Done.

tox_packet/src/dht/crypto_handshake.rs
```diff
@@ -126,7 +130,7 @@ pub struct CryptoHandshakePayload {
     /// used to make sure that possible attacker can't combine payload from old
     /// `CryptoHandshake` with new `Cookie` and try to do mess sending such
     /// packets.
-    pub cookie_hash: sha512::Digest,
+    pub cookie_hash: [u8; <Sha512 as Digest>::OutputSize::USIZE],
```
kpp, line 133: PingId?

kurnevsky: Nope, It's different type which uses sha512.

tox_packet/src/dht/crypto_handshake.rs
```diff
@@ -137,7 +141,7 @@ impl FromBytes for CryptoHandshakePayload {
     named!(from_bytes<CryptoHandshakePayload>, do_parse!(
         base_nonce: call!(Nonce::from_bytes) >>
         session_pk: call!(PublicKey::from_bytes) >>
-        cookie_hash: call!(sha512::Digest::from_bytes) >>
+        cookie_hash: map_opt!(take!(64), |bytes: &[u8]| bytes.try_into().ok()) >>
```
kpp, line 144:
```suggestion
        cookie_hash: map_opt!(take!(sha512::DIGESTBYTES), |bytes: &[u8]| bytes.try_into().ok()) >>
```

kpp (approved):

PR #455, opened 2021-08-21
Pure rust asymmetric crypto

kurnevsky:
libsodium no more


tox_binary_io/src/crypto.rs
```diff
@@ -0,0 +1,52 @@
...
+use super::FromBytes;
+
+
+impl FromBytes for PublicKey {
+    // TODO: use map_from with new nom version
+    named!(from_bytes<PublicKey>, map!(map_opt!(take!(KEY_SIZE), |pk: &[u8]| pk.try_into().ok()), |pk: [u8; KEY_SIZE]| pk.into()));
+}
+
+/* TODO
+Use the following implementation when https://github.com/TokTok/c-toxcore/issues/1169 is fixed.
```
kpp, line 16: It is fixed

tox_binary_io/src/crypto.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use nom::{map, map_opt, named, take};
+
+use crypto_box::{PublicKey, SecretKey, KEY_SIZE};
+
+use super::FromBytes;
+
+
+impl FromBytes for PublicKey {
+    // TODO: use map_from with new nom version
```
kpp, line 11: What's up with new nom version?

tox_binary_io/src/lib.rs
```diff
@@ -140,13 +138,13 @@ macro_rules! unpack {
     ($variable:expr, $variant:path [ $($inner:ident),* ]) => (
         match $variable {
             $variant( $($inner),* ) => ( $($inner),* ),
-            other => panic!("Expected {} but got {:?}", stringify!($variant), other),
```
kpp, line 143: Why did you remove `other` output?

tox_binary_io/src/crypto.rs
kurnevsky: We have outdated nom version, and this function was added only to the latest version.

tox_binary_io/src/lib.rs
kurnevsky: Because it doesn't implement `Debug` anymore (because secret/precomputed keys from rust-crypto don't implement it).

tox_binary_io/src/crypto.rs
kurnevsky: Fixed, but I'd leave it for another MR, because firstly we need to check how many nodes from tox network migrated to a new version.

tox_packet/src/relay/onion_request.rs
```diff
@@ -15,11 +16,11 @@ use nom::combinator::rest;
 
 /// Encrypted payload should contain `IpPort`, `PublicKey` and inner encrypted
 /// payload that should contain at least `IpPort` struct.
-const ONION_MIN_PAYLOAD_SIZE: usize = (SIZE_IPPORT + MACBYTES) * 2 + PUBLICKEYBYTES;
+const ONION_MIN_PAYLOAD_SIZE: usize = (SIZE_IPPORT + <SalsaBox as AeadCore>::TagSize::USIZE) * 2 + crypto_box::KEY_SIZE;
 
 /// `OnionRequest1` packet with encrypted payload from `OnionRequest` packet
 /// shouldn't be bigger than `ONION_MAX_PACKET_SIZE`.
-const ONION_MAX_PAYLOAD_SIZE: usize = ONION_MAX_PACKET_SIZE - (1 + NONCEBYTES + PUBLICKEYBYTES + ONION_RETURN_1_SIZE);
```
kpp, line 22: Do we really have to have NONCEBYTES but not PUBLICKEYBYTES? Why not create an alias to crypto_box::KEY_SIZE?

tox_packet/src/onion/onion_response_3.rs
```diff
@@ -67,7 +66,7 @@ mod tests {
             },
             payload: InnerOnionResponse::OnionAnnounceResponse(OnionAnnounceResponse {
                 sendback_data: 12345,
-                nonce: gen_nonce(),
+                nonce: [42; <SalsaBox as AeadCore>::NonceSize::USIZE],
```
kpp, line 69: There is a const NONCEBYTES

tox_packet/src/messenger/conference/query_response.rs
```diff
@@ -129,47 +129,50 @@ impl PeerInfo {
...
-                temp_pk: gen_keypair().0,
+                real_pk: SecretKey::generate(&mut thread_rng()).public_key(),
+                temp_pk: SecretKey::generate(&mut thread_rng()).public_key(),
                 nickname: "56789".to_owned(),
             }])
     );
 
     #[test]
     fn peer_info_from_bytes_encoding_error() {
+        let mut rng = thread_rng();
```
kpp, line 155: We should either create a tmp var=thread_rng everywhere or use thread_rng() in callsite. See ^ `temp_pk: SecretKey::generate(&mut thread_rng()).public_key(),`

tox_packet/src/relay/onion_request.rs
kpp: To be honest I thought we had tox_crypto crate to be able to keep all these consts in one place.

kurnevsky: > To be honest I thought we had tox_crypto crate to be able to keep all these consts in one place.

It didn't work this way because in tox we have both symmetric and asymmetric crypto, and different kind of hashes. So there are a lot of consts that have same or similar name, but we had here only consts for asymmetric crypto. I'd actually prefer not to redefine consts and import them directly from tox-crypto lib because this way we won't have a confusion - before we used both sodiumoxide and our redefined consts at the same time in different places. As for NONCEBYTES ryst-crypto just don't have a convenient const to use.

tox_packet/src/messenger/conference/query_response.rs
kurnevsky: I used `&mut thread_rng()` inplace sometimes where it's used only once. Everywhere where it's used multiple times it's preferable to cache thread_rng call since it has overhead (not big, but still).

tox_encryptsave/Cargo.toml
```diff
@@ -24,4 +24,4 @@ rand = "0.8"
 sha2 = "0.9"
 xsalsa20poly1305 = "0.7"
 scrypt = "0.7"
-zeroize = "1.4"
+zeroize = "1.3"
```
kpp, line 27: Why?

tox_core/src/state_format/old.rs
```diff
@@ -749,86 +735,4 @@ mod tests {
             },
         ])
     );
-
-    encode_decode_test!(
```
kpp, line 753: Why not keep it?

kpp (approved):
LGTM

tox_encryptsave/Cargo.toml
kurnevsky: crypto_box depends on older version which conflicts with 1.4.

tox_core/src/state_format/old.rs
kurnevsky: encode_decode_test compares result with `Eq`, but it's not implemented for secret and precomputed keys.

PR #457, opened 2021-09-26
refactor: use thiserror instead of failure

### LINKED ISSUE #424: Failure is deprecated
> Notice: failure is deprecated. If you liked failure's API, consider using:
> - Anyhow is a good replacement for `failure::Error`.
> - thiserror is a good, near drop-in replacement for `#[derive(Fail)]`.

kurnevsky:
Fixes #424


tox_core/src/dht/server_ext.rs
```diff
@@ -41,9 +40,8 @@ pub async fn dht_run_socket(
                 Err(e) => {
                     error!("packet receive error = {:?}", e);
                     // ignore packet decode errors
-                    if *e.kind() != DecodeErrorKind::Io { continue }
-                    else {
-                        return Err(Error::new(ErrorKind::Other, e.compat()))
+                    if let DecodeError::Io(e) = e {
+                        return Err(e);
```
kpp, line 44: Continue?

kurnevsky: I inverted the condition - we ignore only decoding errors, not IO.

PR #459, opened 2021-10-31
refactor(nom): update to the latest version

kurnevsky:


kpp:
That's a huge amout of work. Did it affect the binary size? I expect it to shrink a little bit.

kurnevsky:
I didn't check, but it shouldn't affect it much. Ahync/await should be the biggest part of it.

PR #460, opened 2021-11-12
chore(rust): use 2021 edition

kurnevsky:


README.md
```diff
@@ -30,7 +30,7 @@ issue / pull request should be filled on the reference repository.
 [CONTRIBUTING.md](/CONTRIBUTING.md).
 
 ## Building
-Fairly simple. First, install [Rust] >= 1.51.0 and a C compiler ([Build Tools
+Fairly simple. First, install [Rust] >= 1.56.1 and a C compiler ([Build Tools
```
kpp, line 33: Can it be 1.56.0? Also since 1.56 there is a https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field.

kurnevsky: In 1.56.0 they fixed a vulnerability: https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html so it doesn't make sense to advice it.

kpp: Thanks

PR #461, opened 2021-11-13
refactor(dht): introduce a single DHT server state

kurnevsky:
This should fix a deadlock we currently have on master. Later we can
move more stuff in this state and make it pure version of DHT server.


tox_core/src/dht/server/mod.rs
```diff
@@ -233,12 +251,22 @@ impl Server {
 
     /// Check if we have at least one node in good state.
     pub async fn is_connected(&self) -> bool {
-        self.close_nodes.read()
+        self.state.read()
```
kpp, line 254: let state = self.state.read().await;

kpp (approved):
LGTM

PR #464, opened 2022-08-21
refactor(tcp_server): release state lock before sending packets

kurnevsky:
Currently we send tcp packets while holding the state lock. At the
same time we use bounded queues for sending of packets. This means
that a single stuck client can block the whole server until the
timeout happens. This change makes the server release the state lock
before sending any packet.

### DIFF
--- a/tox_core/src/relay/server/server.rs
+++ b/tox_core/src/relay/server/server.rs
@@ -12,11 +12,15 @@ use std::io::{Error, ErrorKind};
 use std::collections::HashMap;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Instant, Duration};
 
-use futures::SinkExt;
+use futures::{SinkExt, StreamExt};
 use futures::channel::mpsc;
 use tokio::sync::RwLock;
+use futures::stream::FuturesUnordered;
+
+/// Interval of time for packet sending
+const TCP_SEND_TIMEOUT: Duration = Duration::from_millis(100);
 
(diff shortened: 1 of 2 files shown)


tox_core/src/relay/server/server.rs
```diff
@@ -42,6 +46,54 @@ struct ServerState {
...
+    while futures.next().await.is_some() { };
+}
+
+enum HandleAction {
+    Send(Vec<(Packet, mpsc::Sender<Packet>)>),
+    Onion(OnionRequest, SocketAddr),
+}
+
+impl Default for HandleAction {
+    fn default() -> Self {
```
kpp, line 72: impl HandleAction::ignore == HandleAction::default and use it instead.

tox_core/src/relay/server/server.rs
```diff
@@ -253,7 +328,7 @@ impl Server {
                 // haven't received DisconnectNotification yet and have sent yet another packet.
                 // In this case we don't want to throw an error and force disconnect the second client.
                 // TODO: failure can be used to return an error and handle it inside ServerProcessor
-                return Ok(())
+                return Ok(HandleAction::default())
```
kpp, line 331: HandleAction::ignore

kpp (changes requested):
So basically this is release state lock + some refactoring. Looks good.

tox_core/src/relay/server/server.rs
kurnevsky: I'd call it `empty` then :)

kpp (approved):

PR #467, opened 2022-11-11
Update deps

kpp:


tox_node/src/node_config.rs
```diff
@@ -36,17 +36,33 @@ impl FromStr for Threads {
...
+/// Specifies where to write logs.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Deserialize)]
+pub enum LogType {
+    Stderr,
+    Stdout,
+    Syslog,
+    None,
+}
+
+impl ValueEnum for LogType {
```
kurnevsky, line 48: Maybe derive this implementation: https://docs.rs/clap/latest/clap/trait.ValueEnum.html

kpp: We can but it’s a +1 feature.

kurnevsky: Yeah, but it's the purpose of this feature, and it can add dependencies only for `tox-node`. You thing we should avoid this? :)

kpp: Why not?

kurnevsky: Derivation adds a bit more safety - if we need to add a new value to `LogType` without derivation we have to remember to add it to `value_variants` manually. Also there could be typos in manually written strings.

kpp: I agree, but it's not something we touch very often.

kurnevsky (approved):

PR #477, opened 2023-11-23
Add functions for tox crate version information

antis81:
Adds functions to query and display the `tox` crate version during runtime.

- `tox::tox_node::crate_version -> String`
- `tox::tox_node::crate_version_major -> u32`
- `tox::tox_node::crate_version_minor -> u32`
- `tox::tox_node::crate_version_patch -> u32`


kpp:
Would you please explain what you are fixing by this PR? Can you add more context? Error messages?

antis81:
Actually this feature is available in c-toxcore ([for example here](https://github.com/TokTok/c-toxcore/blob/24b54722aed65b02f1e921f08f8d31d68f92a45a/toxcore/tox_api.c#L19)). The PR adds those API calls to the "tox" crate. Might be better suited in both "tox" and "tox_core" crate though. This allows to query and display (meta) information about the tox(_core) crate at runtime.

tox/src/lib.rs
```diff
@@ -4,3 +4,39 @@ pub use tox_core as core;
 pub use tox_crypto as crypto;
 pub use tox_encryptsave as encryptsave;
 pub use tox_packet as packet;
+
+pub fn crate_version() -> String {
```
kpp: Missing documentation

tox/src/lib.rs
```diff
@@ -4,3 +4,39 @@ pub use tox_core as core;
...
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn crate_version_is_not_empty() {
+        assert_ne!(crate::crate_version(), "");
+    }
+
+    #[test]
+    fn crate_version_major() {
+        crate::crate_version_major();
```
kpp: At least you could've get crate version from env and compare it.

kpp (changes requested):

kpp:
Ok good. Can you integrate these functions into Tox-node?

antis81:
Okay I moved the version info into the `tox_node` crate together with the version context. Now we have a "human readable" form in addition to this somewhat "cryptic" versioning scheme. However for the time being I say it is fine.

